PR #23, opened 2018-01-09
Expose API error details

thetechnick:


hcloud/error.go
```diff
@@ -11,20 +11,36 @@ const (
...
+}
+
 // IsError returns whether err is an API error with the given error code.
 func IsError(err error, code ErrorCode) bool {
 	apiErr, ok := err.(Error)
 	return ok && apiErr.Code == code
 }
+
+// e := err.(hcloud.Error)
+// details := err.details.(hcloud.InvalidInputDetails)
```
thcyron: Leftovers.

hcloud/schema/error_test.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		if e.Message != "invalid input" {
+			t.Errorf("unexpected Message: %v", e.Message)
+		}
+		if e.Details == nil {
+			t.Fatalf("unexpected Details: %v", e.Details)
+		}
+		if d, ok := e.Details.(*ErrorInvalidInputDetails); !ok {
+			t.Fatalf("unexpected Details type (should be *ErrorInvalidInputDetails): %v", e.Details)
+		} else {
```
thcyron: `else` block not needed since `t.Fatalf` will stop execution.

hcloud/error.go
```diff
@@ -11,20 +11,36 @@ const (
...
+	Details interface{}
 }
 
 func (e Error) Error() string {
 	return fmt.Sprintf("%s (%s)", e.Message, e.Code)
 }
 
+// ErrorInvalidInputDetails contains the details of an 'invalid_input' error.
+type ErrorInvalidInputDetails struct {
+	Fields []*ErrorInvalidInputDetailsField `json:"fields"`
```
thcyron: Structs in root `hcloud` package should not have `json` tags.

thcyron: I would name it `ErrorDetailsInvalidInput` so all details structs will have the same prefix `ErrorDetails`.

thcyron: Also, any particular reason why it’s a pointer and not simply a value? As a rule of thumb, I usually use values for things that don’t have an identity.

hcloud/schema/error.go
```diff
@@ -1,12 +1,43 @@
...
+	}
+	return
 }
 
 // ErrorResponse defines the schema of a response containing an error.
 type ErrorResponse struct {
 	Error Error `json:"error"`
 }
+
+// ErrorInvalidInputDetails defines the schema of the Details field
```
thcyron: Lowercase “details”. 🤷‍♂️

hcloud/schema/error_test.go
thetechnick: the else block was needed because `d` is scoped to the `if` block

PR #24, opened 2018-01-10
Implement datacenter client

thetechnick:


hcloud/datacenter.go
```diff
@@ -1,9 +1,124 @@
...
+	"github.com/hetznercloud/hcloud-go/hcloud/schema"
+)
+
 // Datacenter represents a datacenter in the Hetzner Cloud.
 type Datacenter struct {
 	ID          int
 	Name        string
 	Description string
 	Location    *Location
+	ServerTypes *DatacenterServerTypes
```
thcyron: I would make it a value since it has no identity.

PR #26, opened 2018-01-10
Implement iso client

thetechnick:


hcloud/iso.go
```diff
@@ -18,3 +27,102 @@ const (
 	// ISOTypePrivate is the type of a private ISO image.
 	ISOTypePrivate = "private"
 )
+
+// ISOClient is a client for the iso API.
```
thcyron: Uppercase ISO 😀

hcloud/schema/iso.go
```diff
@@ -7,3 +7,13 @@ type ISO struct {
 	Description string `json:"description"`
 	Type        string `json:"type"`
 }
+
+// ISOGetResponse defines the schema of the response when retrieving a single datacenter.
```
thcyron: datacenter?

hcloud/iso_test.go
```diff
@@ -0,0 +1,212 @@
...
+				ISOs: []schema.ISO{
+					{
+						ID: 1,
+					},
+				},
+			})
+		})
+
+		ctx := context.Background()
+		iso, _, err := env.Client.ISO.GetByName(ctx, "fsn1-dc8")
```
thcyron: Should be a name which makes more sense.

PR #52, opened 2018-01-11
Add pagination, sorting and filtering

### LINKED ISSUE #51: Implement sorting and filtering

thetechnick:
Closes #51

### DIFF
--- a/hcloud/action.go
+++ b/hcloud/action.go
@@ -78,13 +78,48 @@ func (c *ActionClient) GetByID(ctx context.Context, id int) (*Action, *Response,
 	return ActionFromSchema(body.Action), resp, nil
 }
 
+// ActionPage serves as accessor of the actions API pagination.
+type ActionPage struct {
+	Page
+	content []*Action
+}
+
+// Content contains the content of the current page.
+func (p *ActionPage) Content() []*Action {
+	return p.content
+}
+
 // ActionListOpts specifies options for listing actions.
 type ActionListOpts struct {
 	ListOpts
 }
 
-// List returns a list of actions for a specific page.
-func (c *ActionClient) List(ctx context.Context, opts ActionListOpts) ([]*Action, *Response, error) {
+// List returns an accessor to control the actions API pagination.
+func (c *ActionClient) List(ctx context.Context, opts ActionListOpts) *ActionPage {
+	page := &ActionPage{}
+	page.pageGetter = pageGetter(func(start, end int) (resp *Response, exhausted bool, err error) {
+		allActions := []*Action{}
+		if opts.PerPage == 0 {
+			opts.PerPage = 50
+		}
+
+		resp, exhausted, err = c.client.all(func(page int) (*Response, error) {
+			opts.Page = page
+			actions, resp, err := c.list(ctx, opts)
+			if err != nil {
+				return resp, err
+			}
+			allActions = append(allActions, actions...)
+			return resp, nil
+		}, start, end)
+		page.content = allActions
+		return
+	})
+	return page
+}
+
+// list returns a list of actions for a specific page.
+func (c *ActionClient) list(ctx context.Context, opts ActionListOpts) ([]*Action, *Response, error) {
 	path := "/actions?" + valuesForListOpts(opts.ListOpts).Encode()
 	req, err := c.client.NewRequest(ctx, "GET", path, nil)
 	if err != nil {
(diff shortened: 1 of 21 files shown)


hcloud/pagination.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Response returns the Response of the API request.
+func (p *Page) Response() *Response {
+	return p.response
+}
+
+// GoTo fetches the given page.
+// It returns true on success, or false if the resource is exhausted or an error happened while fetching it.
+func (p *Page) GoTo(page int) bool {
```
thcyron: I find it confusing that `GoTo()` actually fetches the page. I was expecting it to just *set* the next page to fetch (i.e. set the cursor) and `Next()` to actually fetch the page.

hcloud/server.go
```diff
@@ -134,14 +134,49 @@ func (c *ServerClient) Get(ctx context.Context, idOrName string) (*Server, *Resp
...
 }
 
+// ServerPage serves as accessor of the servers API pagination.
+type ServerPage struct {
+	Page
+	content []*Server
+}
+
+// Content contains the content of the current page.
+func (p *ServerPage) Content() []*Server {
```
thcyron: Since it’s a specialized struct anyway, what about naming it `Servers()` instead of `Content()`?

hcloud/action.go
```diff
@@ -78,14 +79,70 @@ func (c *ActionClient) GetByID(ctx context.Context, id int) (*Action, *Response,
...
-	path := "/actions?" + valuesForListOpts(opts.ListOpts).Encode()
+// list returns a list of actions for a specific page.
+func (c *ActionClient) list(ctx context.Context, opts ActionListOpts) ([]*Action, *Response, error) {
+	path := "/actions?"
+	if opts.Server != nil {
+		path = fmt.Sprintf("/servers/%d/actions?", opts.Server.ID)
+	}
+	if opts.FloatingIP != nil {
+		path = fmt.Sprintf("/floating_ips/%d/actions?", opts.FloatingIP.ID)
+	}
```
thcyron: I’m not sure if having these filters is a good idea – at least how the work now.

When I specify a sever and a Floating IP, I would expect to get actions concerting both server *and* Floating IP, but right now the Floating IP filter would win and discard the server filter.

I would remove these filters for now.

hcloud/action.go
```diff
@@ -105,23 +162,11 @@ func (c *ActionClient) List(ctx context.Context, opts ActionListOpts) ([]*Action
...
-		if err != nil {
-			return resp, err
-		}
-		allActions = append(allActions, actions...)
-		return resp, nil
-	})
-	if err != nil {
-		return nil, err
+	page := c.List(ctx, opts)
+	if page.All(); page.Err() != nil {
```
thcyron: Looks odd. Why not just use `&&`?

thcyron:
As mentioned in another comment, can we make the `All()` method on a `Page` return content *and* error? It would look like this:

```go
// Get all servers after page 5.
servers, err := client.Server.List(ctx).GoTo(5).All()
```

Additional benefit: we could ditch `client.Server.All()` because `List().All()` would do the same:

```go
// Currently
servers, err := client.Server.All(ctx)
// New
servers, err := client.Server.List(ctx).All()
```

hcloud/action.go
thetechnick: `All` has no return value

hcloud/pagination.go
thetechnick: This would be confusing as well because to get page 6 you would have to call `.GoTo(5)` and then call `.Next()`, so I opted for `GoTo` to fetch the page when run

hcloud/action.go
thcyron: 🤦‍♂️

hcloud/pagination.go
thcyron: Yes, that would be confusing. `Goto(5).Next()` should fetch page 5. Probably `Next` is not a good name.

hcloud/server.go
thetechnick: I think its simpler for the user to use when every Page object has the same method names.

thetechnick:
If we ditch the `.All(ctx)` method this would mean that the user has to pass the list options every time:
```
// New
servers, err := client.Server.List(ctx, opts).All()
```
otherwise I do not see any downsides

thcyron:
I’m still not completely happy with the new interface and I don’t feel comfortable to make such a significant change so close to releasing version 1.0.0. The current interface isn’t bad and we can easily add sorting and filtering to it in a 1.1 release. I hope you understand.

PR #63, opened 2018-01-19
Add Image property to server

thetechnick:
Closes #63.


hcloud/schema/server.go
```diff
@@ -18,6 +18,7 @@ type Server struct {
 	ISO             *ISO            `json:"iso"`
 	Locked          bool            `json:"locked"`
 	Datacenter      Datacenter      `json:"datacenter"`
+	Image           Image           `json:"image"`
```
thcyron: `image` is `nullable`, so should be `*Image`.

PR #64, opened 2018-01-22
Add ActionClient.Watch to watch progress of actions

thetechnick:


hcloud/action.go
```diff
@@ -125,3 +125,62 @@ func (c *ActionClient) All(ctx context.Context) ([]*Action, error) {
 
 	return allActions, nil
 }
+
+// Wait watches the action state until it completes with success or error.
+// Also provides progress updates on the first channel.
+func (c *ActionClient) Wait(ctx context.Context, action *Action) (<-chan int, <-chan error) {
```
thcyron: I don’t think Wait is a good name because it does not wait. Maybe Watch or WatchProgress?

hcloud/action.go
```diff
@@ -125,3 +125,62 @@ func (c *ActionClient) All(ctx context.Context) ([]*Action, error) {
...
+		sendProgress := func(p int) {
+			select {
+			case progressCh <- p:
+				break
+			default:
+				break
+			}
+		}
+
+		var retries = 0
```
thcyron: What about resetting it to 0 after a successful response?

hcloud/action_test.go
```diff
@@ -134,3 +134,76 @@ func TestActionClientAll(t *testing.T) {
...
+		select {
+		case progress := <-progressCh:
+			progressUpdates = append(progressUpdates, progress)
+		case err = <-errCh:
+			break loop
+		}
+	}
+
+	if err == nil {
+		t.Fatalf("error expected %v", err)
```
thcyron: Wrong message?

hcloud/action_test.go
```diff
@@ -134,3 +134,76 @@ func TestActionClientAll(t *testing.T) {
...
+	}
+	if len(progressUpdates) != 1 && progressUpdates[0] != 50 {
```
thcyron: Should be `||`?

hcloud/action.go
thetechnick: done

hcloud/action_test.go
thetechnick: Is correct because i expect an non nil `err` because the state of the action is error

thcyron: Your condition might be correct, but why would you print `error expected <nil>`?

PR #72, opened 2018-02-02
Provide a godoc conform example

mxschmitt:
This PR includes a new example, so that godoc will recognize it and renders it correctly. 
Of course this unit test will fail, but per default it won't be executed, so everything should be fine. Maybe a good idea could be to extend it with some more functionality.


thcyron:
Didn’t know you can do that. Thank you!

PR #78, opened 2018-03-09
Allow specifying SSH key by fingerprint

### LINKED ISSUE #77: Allow SSHKey fingerprint instead of SSHKey ID
While creating an instance it could be nice to use the SSHKey fingerprint instead of the primary key of an SSHKey. Looks like that API already supports that as an undocumented feature.

I would like to use that for a Hetzner Cloud integration to avoid another roundtrip to get the ID for a key instead of simply providing the fingerprint.

thcyron:
Closes #77.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes
 
+## master
+
+* Allow specifying SSH key via fingerprint when creating a server
+  ([issue #77](https://github.com/hetznercloud/hcloud-go/issues/77))
+
 ## v1.4.0
 
 * Retry all calls that triggered the API ratelimit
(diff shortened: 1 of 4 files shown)


thcyron:
After internal discussion we decided to not implement this feature. The API supports this, but it’s undocumented and will be removed from the API soon.

The reasoning for this is that we might want to support specifying SSH keys by name in the future (as they are unique per project). Both fingerprint and name are strings and we wouldn’t be able to decide whether to lookup the SSH key by fingerprint or name.

tboerger:
Is there any ETA when it will be possible to get it by name? Beside that, the index route already supports searching by name, or not?

thcyron:
There is not ETA when specifying SSH keys by name when creating a server will be possible.

The `/ssh_keys` API endpoint supports a `name` query string argument to retrieve an SSH key by its name. Use this call to get the ID and use the ID in the create server call.

PR #97, opened 2018-09-26
Remove the possibility to specify the backup window

LKaemmerling:


thcyron:
Do you think it’s a good idea to change the signature of the `EnableBackup()` method and thus making a backwards-incompatible change? I know that removing the possibility to specify the backup window is backwards-incompatible by itself, but there are two ways to handle that in the client:

1. Remove the window parameter completely from the code. Users using the new version will get an error at compile-time.

2. Leave the window parameter, but when it is set, immediately return an error. Will not break users at compile-time, but at runtime.

Solution 1 violates semantic versioning, but users will notice immediately that specifying the backup window is no longer supported.

Solution 2 does not violate semantic versioning, but relies on users doing proper error handling when calling `EnableBackup()` (which they should do anyway, right?).

Once we return an error in the API when a backup window is specified, old clients will get an error nevertheless, i.e. it will behave just like solution 2.

Regardless of what solution we will implement, we can’t merge that right away.

LKaemmerling:
I think a compile-time break is easier to fix than a runtime error. The decision was to break the API "hard," so we should do that on hcloud-go as well. And one more point: sharing on all other hcloud-go makes validating the input of our API, not hcloud-go, so we should keep this here.

thcyron:
We decided to not make source code changes introducing compile-time errors.

> And one more point: sharing on all other hcloud-go makes validating the input of our API, not hcloud-go, so we should keep this here.

Right. So we make no changes to the client and let the API return an error.

I close this issue then because there’s nothing to do right now. Later, when the API returns an error when a backup window is specified, we can add a deprecation notice to the `EnableBackup()` documentation.

PR #101, opened 2018-11-01
Fixed PublicNet IPv6 extraction from returned network

CuBiC3D:
I experienced some problems setting up IPv6 DNS entries using Terraform due to invalid `${hcloud_server.gitlab.ipv6_address}` variable. Hetzner Cloud returns the following according to the API documentation(see https://docs.hetzner.cloud/#servers-get-all-servers) for IPv6 on server requests:
```
"ipv6": {
          "ip": "2001:db8::/64",
          "blocked": false,
          "dns_ptr": [
            {
              "ip": "2001:db8::1",
              "dns_ptr": "server.example.com"
            }
          ]
        }
```
Using go's net.ParseCIDR returns `2001:db8::` for the IP part resulting `${hcloud_server.gitlab.ipv6_address}` in Terraform being not a valid first IPv6 address.

I'm not sure if this workaround is an appropriate fix.

### DIFF
--- a/hcloud/schema.go
+++ b/hcloud/schema.go
@@ -186,6 +186,9 @@ func ServerPublicNetIPv6FromSchema(s schema.ServerPublicNetIPv6) ServerPublicNet
 		DNSPtr:  map[string]string{},
 	}
 	ipv6.IP, ipv6.Network, _ = net.ParseCIDR(s.IP)
+	// Workaround to extract the first IP from IPv6 range.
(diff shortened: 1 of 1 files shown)


thcyron:
To me, this looks like an issue in the Terraform provider itself. The provider sets `ipv6_address` and `ipv6_network`, and the former should be set to the first IPv6 address which is configured on the host (i.e. network + 1).

I created an issue in the provider’s repository: https://github.com/terraform-providers/terraform-provider-hcloud/issues/39

PR #120, opened 2019-09-16
Add support for names to Floating IPs

LKaemmerling:


hcloud/floating_ip.go
```diff
@@ -165,6 +191,9 @@ func (c *FloatingIPClient) Create(ctx context.Context, opts FloatingIPCreateOpts
 		Type:        string(opts.Type),
 		Description: opts.Description,
 	}
+	if opts.Name != nil {
```
thcyron: No need to check for non-nil here, just like description.

thcyron:
Please add a changelog line.

PR #130, opened 2020-04-23
Implemented DnsServer api-client for managing records and zones.

ptr1120:
Hello,
it would be nice to integrate this for managing Hetzner Dns.

Regards,
Peter

### DIFF
--- a/hcloud/client.go
+++ b/hcloud/client.go
@@ -70,6 +70,7 @@ type Client struct {
 	ServerType ServerTypeClient
 	SSHKey     SSHKeyClient
 	Volume     VolumeClient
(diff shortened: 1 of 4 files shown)


LKaemmerling:
Hello Peter,

Thanks for your contribution!

Unfortunately, this library is the wrong place for the DNS API Integration.
In `hcloud-go` we will only implement services that are accessible directly via the Hetzner Cloud API (https://docs.hetzner.cloud).

You should see clients/libraries for the DNS API show up in the next few weeks, I know a few people are working on this topic.

I wish you all the bests,
Lukas

PR #132, opened 2020-05-08
Use http.DefaultClient by default

thcyron:
As stated in the net/http docs, a http.Client should be reused in order
to take advantage of cached TCP connections. http.DefaultClient is safe
for concurrent use.

This will improve performance when you create a lot of clients with
different tokens, for example.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 * Add `Status` field to `Volume`
 * Add subnet type `cloud`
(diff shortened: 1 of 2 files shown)


LKaemmerling:
As we talked about this, I will close this MR. When we use the default client now this can lead to different behavior of the hcloud-go sdk. Therefore we will close this and just merge #133 .

PR #137, opened 2020-06-24
Make GetByName() return nil for empty names

### LINKED ISSUE #136: I deleted the wrong server because empty server arg selects the first server (?)
Feel free to send me to the hcloud cli github, but I feel the problem lies in the API.

command:
hcloud server delete "$( hcloud server list | awk '$2 == "already-manually-deleted-server.domain.net" {print $1}' )"
In effect: hcloud delete ""

Deletes the first server (order determined by hcloud server list).

Why would I do that? I wanted to use the server I payed for for the month before deleting it, so I put this delete command into the crontab for June 22 0:00. In the meantime, someone (me?) deleted it manually, so the list command returned an empty string "" so that hcloud delete "" happened.

Whatever convenience you had in mind, this convenience stops when you can delete the wrong server by this or a typo.

In the mean time, I learned that you can use the server name, too, so the awk command was unnecessary. As an aside, can you make that clear in the help output?
hcloud server describe --help
Describe a server

Usage:
  hcloud server describe [FLAGS] SERVER

Flags:
  -h, --help                 help for describe
  -o, --output stringArray   output options: json|format

Global Flags:
      --poll-interval duration   Interval at which to poll information, for example action progress (default 500ms)

ADD: SERVER can refer to id or name

As my nick name suggests, I would accept if you just decline this request.

E: I got the server back via support (apparantly that is possible within *up to* 24h).

E2: The code is straightforward and starts here: https://github.com/hetznercloud/cli/blob/50a7de3c37155fd71ff1e6cb0f9206eeb6c37eb8/cli/server_delete.go#L23

The code behind the API request (/servers?name=[empty string]) is not open source, I suppose.
Documentation says "Can be used to filter Servers by their name. The response will only contain the Server matching the specified name." well unless it is empty, then it shows all.

thcyron:
In CertificateListOpts, Name is a string, thus we cannot differentiate
between Name unset and Name set to empty string. Making it a *string
would be a breaking change. So the only thing we can do is to
explicitly catch empty name in all those GetByName() methods.

Fixes #136.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## master
+
+* Make all `GetByName` methods return `nil` when an empty name is provided
+
 ## v1.18.0
 
 * Add `Status` field to `Volume`
--- a/hcloud/certificate.go
+++ b/hcloud/certificate.go
@@ -51,6 +51,9 @@ func (c *CertificateClient) GetByID(ctx context.Context, id int) (*Certificate, 
 
 // GetByName retrieves a Certificate by its name. If the Certificate does not exist, nil is returned.
 func (c *CertificateClient) GetByName(ctx context.Context, name string) (*Certificate, *Response, error) {
+	if name == "" {
+		return nil, nil, nil
+	}
 	Certificate, response, err := c.List(ctx, CertificateListOpts{Name: name})
 	if len(Certificate) == 0 {
 		return nil, response, err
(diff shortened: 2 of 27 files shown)


LKaemmerling:
Would it make sense to return an Error instead of returning just nil? I guess this would point the hcloud-go user to a better direction than just having everything nil.

thcyron:
When you query for a server by ID and that server doesn’t exist, we return nil, not an error. I just went the same route.

What would be the advantage of returning `errors.New("hcloud: invalid server name")` instead?

PR #141, opened 2020-07-14
use DumpRequestOut for debugging

pschultz:
httputil.DumpRequest was used before, which should only be used by
servers. Clients should use httputil.DumpRequestOut, which may produce
subtly different (and more accurate) output.

> DumpRequestOut is like DumpRequest but for outgoing client requests. It includes any headers that the standard http.Transport adds, such as User-Agent.
-- https://golang.org/pkg/net/http/httputil/#DumpRequestOut

I don't think it's worth mentioning this in the changelog. If you disagree let me know and I'll update it.


thcyron:
Thank you!

PR #193, opened 2022-02-19
Fix metadata client error detection

choffmeister:
Fixes two issues:

1. Metadata client does not recognise unsuccessful HTTP response codes
2. Closed a potential memory leak (not closed HTTP response)


LKaemmerling:
Hey @choffmeister,

thank you good catch!

PR #194, opened 2022-03-14
Catch invalid token values and error out without value exposure

NotTheEvilOne:
This PR handles tokens that have been previously for example encoded in base64 for Kubernetes in an erroneous way containing newline characters.

Currently the error returned return the token value. This may be an unexpected value exposure not intended by an operator.

Example:
`net/http: invalid header field value "Bearer broken\n" for key Authorization`

Applying this change changes the error to not contain any sensitive data anymore:
`Authorization token contains invalid characters`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/prometheus/client_golang v1.11.0
 	github.com/stretchr/testify v1.7.0
+	golang.org/x/net v0.0.0-20200625001655-4c5254603344
 )
 
 require (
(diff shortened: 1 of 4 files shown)


NotTheEvilOne:
Hi,

> Would it be possible that you add a test case as well?
I'll do that.

> Also we would suggest mirroring the function `httpguts.ValidHeaderFieldValue(token)` into an internal function, because we would like to not include more dependencies. Would this be possible for you?

I'm not sure if that makes a difference. In fact `net/http` uses the [same dependency](https://github.com/golang/go/blob/master/src/net/http/header.go#L197) already. The dependency and code use is already given therefore and would only result in code duplication. Please advice if you would like to integrate the function anyway.

LKaemmerling (approved):

LKaemmerling:
Hey @NotTheEvilOne,

thank you, you are right!

PR #198, opened 2022-05-09
Set UsePrivateIP for targets when creating a LoadBalancer

hakman:
Setting `UsePrivateIP` for targets when creating a LoadBalancer should be allowed, same as when creating the targets separately:
https://github.com/hetznercloud/hcloud-go/blob/cf8f35d1f96cc8d74d1706f9e834ed25a851300f/hcloud/load_balancer.go#L567-L573
https://github.com/hetznercloud/hcloud-go/blob/cf8f35d1f96cc8d74d1706f9e834ed25a851300f/hcloud/load_balancer.go#L597-L603


LKaemmerling:
Hey @hakman,

thank you! Good catch!

PR #199, opened 2022-05-13
fix(images): Fix inabiltiy to find depreacted images by name

SISheogorath:
Currently it's not possible to query e.g. the `fedora-35` image by name,
after it was marked as deprecated due to the release of the `fedora-36`
image. Instead the library will treat this request as if the image
doesn't exist anymore.

This patch enables the `IncludeDeprecated` switch in the `GetByName`
function to allow images that were marked as deprecated to be found by
name.

This is necessary to allow e.g. the Terraform module, build on top of
this library, to continue to provide old OS versions for a transition
time without breaking.

References:
https://github.com/hetznercloud/terraform-provider-hcloud/blob/1dbb289398f19ca52b6b772501630381fc601fa0/internal/server/resource.go#L230
https://github.com/hetznercloud/terraform-provider-hcloud/issues/527

### DIFF
--- a/hcloud/image.go
+++ b/hcloud/image.go
@@ -102,7 +102,7 @@ func (c *ImageClient) GetByName(ctx context.Context, name string) (*Image, *Resp
 	if name == "" {
 		return nil, nil, nil
 	}
-	images, response, err := c.List(ctx, ImageListOpts{Name: name})
+	images, response, err := c.List(ctx, ImageListOpts{Name: name, IncludeDeprecated: true})
 	if len(images) == 0 {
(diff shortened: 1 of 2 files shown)


LKaemmerling:
Hey @SISheogorath,

sorry for the late reply. Imho this change is quite "drastic". Because now, all deprecated images will be silently "available" again. So as a sample how would you notice it in terraform? On our cli we have a special flag `--allow-deprecated-image` maybe on Terraform we should do something similar. Therefore I would prefer to reject this change here and would like to move it to Terraform.

PR #209, opened 2022-09-04
Add delete server response

jeremylevy:
Hello 👋,

I'm trying to find a way to wait until a server is deleted (i.e. unreachable).

The API documentation states that an action object is returned when deleting a server (https://docs.hetzner.cloud/#servers-delete-a-server) but this object is absent from the response of the server's delete method.

It's a breaking change given that the number of values returned needs to be changed.

### DIFF
--- a/hcloud/schema/server.go
+++ b/hcloud/schema/server.go
@@ -136,6 +136,11 @@ type ServerCreateResponse struct {
 	NextActions  []Action `json:"next_actions"`
 }
 
+// ServerDeleteResponse defines the schema of the response when deleting a server.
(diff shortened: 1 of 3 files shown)


LKaemmerling:
Hey @jeremylevy,
Thank you, this is a good idea! However I would like to avoid breaking changes as this would include the need of doing a v2 which we want to avoid with the hcloud go as we want to have the same "major" version of hcloud go together with the API simply to make it easier to understand:)

jeremylevy:
Ok, no problem.

I will try to find a way to access the action object directly from the response body to solve my issue.

Thanks Lukas!

PR #210, opened 2022-09-19
Stop automatic retrying on RateLimitExceeded

LKaemmerling:
It does not make sense to retry when the rate limit is already exceeded. However, it makes sense to retry on Conflict. In the best case this would happen on the Caller/User side, but removing the retry completely would be a breaking change.

Signed-off-by: Lukas Kämmerling <lukas.kaemmerling@hetzner-cloud.de>


4ND3R50N (approved):
LGTM

PR #219, opened 2022-12-19
feat: Use generics to get pointers to types

daenney:
Since the minimum required Go version is 1.18 according to go.mod we can use generics to have a single function that can return a pointer to any type. This avoids the need to add more $Type helpers in the future.

The new Ptr function is introduced and exsiting functions are updated to call through Ptr instead. Any exported functions are now marked as deprecated but not removed in order to avoid breaking backwards compatibility. All non-exported functions are removed. All call sites are updated to use Ptr directly.


apricote:
Thanks for the contribution @daenney!

We still test against Go 1.17, even though the `go.mod` specifies v1.18+.

We should remove the tests for Go v1.17 and codify our Go version requirements (probably in line with [official support](https://go.dev/doc/devel/release#policy)).

I will create a PR for this and notify you once that is merged :)

apricote:
#220 is merged, you may now rebase your branch against `main`.

daenney:
@apricote Thanks! Pushed a rebase.

PR #222, opened 2022-12-25
Fix schema for primary IP pricing to reflect the current API response

byReqz:
The /pricing API endpoint for primary IPs does not return individual DCs anymore but just regions ("fsn1", "nbg1", ...). I have changed the key name in the schema and updated the test for the new format.

### DIFF
--- a/hcloud/schema/pricing.go
+++ b/hcloud/schema/pricing.go
@@ -97,7 +97,8 @@ type PricingGetResponse struct {
 // PricingPrimaryIPTypePrice defines the schema of pricing information for a primary IP
 // type at a datacenter.
 type PricingPrimaryIPTypePrice struct {
-	Datacenter   string `json:"datacenter"`
+	Datacenter   string `json:"datacenter"` // Deprecated: the API does not return pricing for the individual DCs anymore
+	Location     string `json:"location"`
 	PriceHourly  Price  `json:"price_hourly"`
 	PriceMonthly Price  `json:"price_monthly"`
 }
--- a/hcloud/pricing.go
+++ b/hcloud/pricing.go
@@ -56,7 +56,8 @@ type FloatingIPTypePricing struct {
 // PrimaryIPTypePricing defines the schema of pricing information for a primary IP
 // type at a datacenter.
 type PrimaryIPTypePricing struct {
-	Datacenter string
+	Datacenter string // Deprecated: the API does not return pricing for the individual DCs anymore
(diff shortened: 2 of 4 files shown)


hcloud/schema/pricing.go
```diff
@@ -97,7 +97,7 @@ type PricingGetResponse struct {
 // PricingPrimaryIPTypePrice defines the schema of pricing information for a primary IP
 // type at a datacenter.
 type PricingPrimaryIPTypePrice struct {
-	Datacenter   string `json:"datacenter"`
+	Datacenter   string `json:"location"`
```
jangraefen, line 100: I think it is a bit suprising the get a location, while the attribute is named `Datacenter`. I get that renaming the attribute would break downward compatibilities, but existing code might also break if it now gets a location instead of a datacenter.

If downward compatibility should be preserved, my suggestion would be to add a second attribute called `Location`. Then add a comment to `Datacenter`, marking it as deprecated, if its not provided by the API.

apricote:
Looks like we changed the API after implementing it in `hcloud-go`. I will try to make sure that such things do not happen in the future.

hcloud/schema/pricing.go
apricote: I agree with @jangraefen.

To make this compatible with tooling, the comment should start with `Deprecated: ...`.

Please also make sure to add the new field to `hcloud.PrimaryIPPricing` and `hcloud.PricingFromSchema()`.

byReqz: I have added the separate key and deprecated `Datacenter`.

apricote:
Thanks for the fix @byReqz!

jangraefen:
Hey @apricote,

are you planning on releasing the fix in the near future or are you guys waiting for other things? In case the release will take some time, I am considering disabling the primary IP address future from my Prometheus exporter for now, to prevent any further crashes or issues for my users. If a release is imminent though,  I think a temorary release for the Prometheus exporter would be overkill though 😉.

With kind regards,
Jan

apricote:
I can publish a release tomorrow :)

jangraefen:
@apricote: Wow, thats a quick response and a nice one at that! Thanks for being so quick and for the good work! 🙂

apricote:
@jangraefen Release v1.39.0 is published with these changes. https://github.com/hetznercloud/hcloud-go/releases/tag/v1.39.0

PR #268, opened 2023-06-05
feat(servertype): implement new Deprecation api field

apricote:
We recently added a new field to the ServerType API response that indicates if and when the server type was deprecated. The same structure will also be used for other resources in the future, replacing the current `deprecated` fields.

Marked as draft until I have finished the implementation in our other Go Integrations, to make sure that this feels ergonomic.


DrBu7cher (approved):
:+1:

PR #273, opened 2023-06-16
fix: adjust label validation for max length of 63 characters

fsrv-xyz:
According to the api documentation, the length of a label value must be a string of 63 characters or fewer.
The currently used regex for the validation allows 64 char long values.

### DIFF
--- a/hcloud/labels.go
+++ b/hcloud/labels.go
@@ -6,8 +6,8 @@ import (
 )
 
 var keyRegexp = regexp.MustCompile(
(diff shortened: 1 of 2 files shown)


apricote:
Thanks a lot @fsrv-xyz! We should also make the same fixes in [`hcloud-python`](https://github.com/hetznercloud/hcloud-python/blob/47eb9f1c79e05a61084f0a639f9497beb22d6910/hcloud/helpers/labels.py#L6-L11)

PR #277, opened 2023-06-21
feat(network): add field expose_routes_to_vswitch

apricote:
Feature was released today.

Marked as draft until the other Go integrations are ready.

### DIFF
--- a/hcloud/network.go
+++ b/hcloud/network.go
@@ -45,6 +45,9 @@ type Network struct {
 	Servers    []*Server
 	Protection NetworkProtection
 	Labels     map[string]string
+
+	// ExposeRoutesToVSwitch indicates if the routes from this network should be exposed to the vSwitch connection.
+	ExposeRoutesToVSwitch bool
 }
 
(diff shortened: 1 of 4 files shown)


hcloud/network.go
```diff
@@ -226,6 +236,9 @@ type NetworkCreateOpts struct {
 	Subnets []NetworkSubnet
 	Routes  []NetworkRoute
 	Labels  map[string]string
+	// ExposeRoutesToVSwitch indicates if the routes from this network should be exposed to the vSwitch connection.
+	// The exposing only takes effect if a vSwitch connection is active.
+	ExposeRoutesToVSwitch bool
```
jooola, line 241: Shouldn't this field be optional when creating a network ? And have nil instead of false ?

apricote: Its optional, if you don't specify it, go will default this to `false`, which is also the default from the API.

I think not having a pointer here is more ergonomic for the user, otherwise they would need to pass `ExposeRoutesToVSwitch: hcloud.Ptr(true)`

jooola (approved):

PR #280, opened 2023-06-26
docs: fix link to deprecated reference

jooola:


hcloud/iso.go
```diff
@@ -89,7 +89,7 @@ type ISOListOpts struct {
 	Architecture []Architecture
 	// IncludeWildcardArchitecture must be set to also return custom ISOs that have no architecture set, if you are
 	// also setting the Architecture field.
-	// Deprecated: Use the IncludeArchitectureWildcard field instead.
+	// Deprecated: Use the [IncludeArchitectureWildcard] field instead.
```
apricote, line 92:
```suggestion
	// Deprecated: Use [ISOListOpts.IncludeArchitectureWildcard] instead.
```

I tried it locally with only the field name and that did not properly setup the link in my IDE. I can confirm that its working if you add the full path.

apricote (approved):

Issue #263, opened 2023-05-15
Breaking Change: ID fields will become int64


apricote:
## 52-bit IDs

Our API documentation specified the type for all ID fields as `integer`. We have realized that this was not sufficient and that some users of our API (like our own `hcloud-go`) assumed that 32-bit integers would be enough. Soon we will have IDs with up to 52 bits, and this can cause issues if your code does not handle this properly. Please make sure that your integration with our API uses 64-bit integers (also called `long` in many languages) for all ID fields, and fields that contain IDs such as `network.servers`.
    
We will start using the larger IDs after **September 1st 2023**.
    
To comply with this, we plan on releasing a new major version of our own `hcloud-go` library to switch the data type for IDs from `int` to `int64` in the next weeks. We will maintain the existing version as well as the new version for the migration period.

**Important**: To avoid any issue, you should make sure that all necessary changes are already completed by the time mentioned above.

apricote:
Version `2.0.0` was just published for these changes. Please make sure to upgrade your usage before **September 1st 2023** to avoid any issues.

Until then, we will keep backporting changes to 1.x and releasing new versions.

Srakbi:
> ## 52-bit IDs
> 
> Our API documentation specified the type for all ID fields as `integer`. We have realized that this was not sufficient and that some users of our API (like our own `hcloud-go`) assumed that 32-bit integers would be enough. Soon we will have IDs with up to 52 bits, and this can cause issues if your code does not handle this properly. Please make sure that your integration with our API uses 64-bit integers (also called `long` in many languages) for all ID fields, and fields that contain IDs such as `network.servers`.
>     
> We will start using the larger IDs after **September 1st 2023**.
>     
> To comply with this, we plan on releasing a new major version of our own `hcloud-go` library to switch the data type for IDs from `int` to `int64` in the next weeks. We will maintain the existing version as well as the new version for the migration period.
> 
> **Important**: To avoid any issue, you should make sure that all necessary changes are already completed by the time mentioned a

jooola:
The v1 maintenance branch is not supported anymore.

See #590.

PR #282, opened 2023-07-05
feat!: use int64 for ID fields

apricote:
The `int` datatype is not guaranteed to offer 64 bits of precision. On 32-bit platforms it only has 32 bits of precision. This is incompatible with the Hetzner Cloud identifiers having up to 52-bits after September 1st 2023.

See #263 for more details about this change.


apricote:
- Update in cli: hetznercloud/cli#512

jooola (approved):

PR #293, opened 2023-07-20
feat(metadata): add timeout option

apricote:
Some users might try to access the metadata service even though their networking setup breaks connectivity. This might lead to forever stalled connections. By setting a timeout we can error out instead of waiting indefinitely.

In my tests the metadata service responded in 12-40ms, so a default timeout of 1 second should be enough to only fail when the metadata service is actually unavailable.

### DIFF
--- a/hcloud/metadata/client.go
+++ b/hcloud/metadata/client.go
@@ -49,7 +49,7 @@ func WithInstrumentation(registry *prometheus.Registry) ClientOption {
 	}
 }
 
-// WithTimeout specifies a time limit for requests made by this [Client].
+// WithTimeout specifies a time limit for requests made by this [Client]. Defaults to 5 seconds.
 func WithTimeout(timeout time.Duration) ClientOption {
 	return func(client *Client) {
 		client.timeout = timeout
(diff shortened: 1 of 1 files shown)


hcloud/metadata/client.go
```diff
@@ -18,55 +19,65 @@ const Endpoint = "http://169.254.169.254/hetzner/v1/metadata"
...
+		client.timeout = timeout
+	}
+}
+
+// NewClient creates a new [Client] with the options applied.
 func NewClient(options ...ClientOption) *Client {
 	client := &Client{
 		endpoint:   Endpoint,
 		httpClient: &http.Client{},
+		timeout:    1 * time.Second,
```
jooola, line 64: I think 1s as default timeout value is too aggressive, we lose the flexibility of automatically recovering if the network is having some issue, because the request would time out before that.

I would either keep the behavior as it was before, no default timeout, or at least raise it to 5/10/15 seconds.

apricote: The metadata service is running on each physical host system. There should be no excessive networking between the VM and it, it all happens on the same hardware.

5 seconds sounds good to me :)

jooola (approved):

PR #304, opened 2023-08-15
fix: include running action progress in WatchOverallProgress

jooola:
Include the running action progress in the actions progress calculation.

The function now considers a failed action as completed, and is included in the progress. Not counting the failed action as completed might send some weird progress values e.g. [0%, 75%, (failed action occurs here) 50%], while if we count them, we should have [0%, 75%, 100%].

Related to https://github.com/hetznercloud/cli/issues/528

### DIFF
--- a/hcloud/action.go
+++ b/hcloud/action.go
@@ -189,13 +189,14 @@ func (c *ActionClient) WatchOverallProgress(ctx context.Context, actions []*Acti
 		defer close(errCh)
 		defer close(progressCh)
 
-		successIDs := make([]int64, 0, len(actions))
+		completedIDs := make([]int64, 0, len(actions))
 		watchIDs := make(map[int64]struct{}, len(actions))
 		for _, action := range actions {
 			watchIDs[action.ID] = struct{}{}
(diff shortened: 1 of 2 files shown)


hcloud/action.go
```diff
@@ -224,20 +225,27 @@ func (c *ActionClient) WatchOverallProgress(ctx context.Context, actions []*Acti
...
 				case ActionStatusError:
 					delete(watchIDs, a.ID)
+					completedIDs = append(completedIDs, a.ID)
 					errCh <- fmt.Errorf("action %d failed: %w", a.ID, a.Error())
 				}
 			}
 
+			progress += (len(completedIDs) * 100)
+			if progress != 0 && progress != previousProgress {
+				sendProgress(progressCh, int(progress/len(actions)))
```
apricote, line 245: `progress / len(actions)` has the potential to panic if the user passes 0 actions to this method. Not an issue right now because we have a check `len(as)==0 { return }` but good to be mindful of this.

apricote (approved):

PR #309, opened 2023-08-24
feat: support resource-specific Action endpoints (#295)

apricote:
Related to https://docs.hetzner.cloud/changelog#2023-06-29-resource-action-endpoints

The existing ActionClient has two purposes:

- Expose the GET /v1/actions & GET /v1/actions/:id endpoints
- Provider helper methods for waiting on running actions

The new resource-specific endpoints do not need the helper methods, so I have refactored the code for the endpoints into the ResourceActionClient. This client is exposed by all resources that have actions as client.Server.Action. In addition, I have replaced the implementation of the two endpoints in ActionClient by forwarding them to an instance of ResourceActionClient.

(cherry picked from commit ddc2ac45489c48a7563806c425222236ab1f8aa0 #295)

### DIFF
--- a/hcloud/action.go
+++ b/hcloud/action.go
@@ -72,25 +72,12 @@ func (a *Action) Error() error {
 
 // ActionClient is a client for the actions API.
 type ActionClient struct {
-	client *Client
+	action *ResourceActionClient
 }
 
 // GetByID retrieves an action by its ID. If the action does not exist, nil is returned.
 func (c *ActionClient) GetByID(ctx context.Context, id int) (*Action, *Response, error) {
-	req, err := c.client.NewRequest(ctx, "GET", fmt.Sprintf("/actions/%d", id), nil)
-	if err != nil {
-		return nil, nil, err
-	}
-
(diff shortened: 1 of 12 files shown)


hcloud/action.go
```diff
@@ -241,9 +198,9 @@ func (c *ActionClient) WatchOverallProgress(ctx context.Context, actions []*Acti
 				}
 			}
 
-			progress += (len(completedIDs) * 100)
+			progress += len(completedIDs) * 100
 			if progress != 0 && progress != previousProgress {
-				sendProgress(progressCh, int(progress/len(actions)))
+				sendProgress(progressCh, progress/len(actions))
```
jooola, line 203: Why are those changes here ?

apricote: Because they are part of the backported PR: https://github.com/hetznercloud/hcloud-go/pull/295/files#diff-8914d399a5e4a0ca8ac89a130609df3ba4006e707596cdbff04855cf929b2dcdR200-R202

PR #320, opened 2023-09-19
feat(error): include http response in api errors

apricote:
This allows us to show more useful information in debug logs in downstream projects.

### DIFF
--- a/hcloud/client.go
+++ b/hcloud/client.go
@@ -286,8 +286,8 @@ func (c *Client) Do(r *http.Request, v interface{}) (*Response, error) {
 			return response, fmt.Errorf("hcloud: error reading response meta data: %s", err)
 		}
 
-		if resp.StatusCode >= 400 && resp.StatusCode <= 599 {
-			err = errorFromResponse(resp, body)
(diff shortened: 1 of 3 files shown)


hcloud/client.go
```diff
@@ -371,7 +371,10 @@ func errorFromResponse(resp *http.Response, body []byte) error {
 	if respBody.Error.Code == "" && respBody.Error.Message == "" {
 		return nil
 	}
-	return ErrorFromSchema(respBody.Error)
+
+	hcErr := ErrorFromSchema(respBody.Error)
+	hcErr.HTTPStatusCode = resp.StatusCode
+	return hcErr
```
jooola, line 377: Oh, this one fell off my radar.

I am not sure if it is smart to diverge from the API schema here, in particular if we use a field.

What do you think about adding a `hcErr.Response()` method that returns the embedded `hcErr.response` value.

apricote: Nice suggestion, makes it more versatile.

I have updated your commit to work with our own `http.Response` wrapper.

jooola (approved):

PR #342, opened 2023-11-15
feat: add interfaces for client structs

pauhull:
This PR adds auto-generated interfaces for client structs. This allows easier mocking of clients for tests.
The interfaces will have the same name as the corresponding struct, except for beginning with an `I`. So for example, the interface for `ServerClient` would be `IServerClient`.
The SDK itself will not use these new interface types, as most client structs have exported properties and using the interface types would thus be a breaking change.


apricote (commented):
Very Nice! Before merging, could you update `cli` to use these new interfaces and open a draft MR? Just to verify that this works as expected.

PR #348, opened 2023-12-06
fix: make schemas consistent with API

pauhull:
In some places the schema definitions are not consistent with the API documentation, for example properties may not have pointer receivers even though they are marked as optional in the documentation. This results in inconsistent JSON when converting `JSON -> Schema -> JSON` (as seen for example in https://github.com/hetznercloud/cli/issues/461, where one output is the exact API response and one is serialized and then converted back to JSON).

This PR aims to fix this issue by making the schema definitions consistent.

### DIFF
--- a/hcloud/schema/image.go
+++ b/hcloud/schema/image.go
@@ -11,7 +11,7 @@ type Image struct {
 	Description  string            `json:"description"`
 	ImageSize    *float32          `json:"image_size"`
 	DiskSize     float32           `json:"disk_size"`
-	Created      time.Time         `json:"created"`
+	Created      *time.Time        `json:"created"`
 	CreatedFrom  *ImageCreatedFrom `json:"created_from"`
 	BoundTo      *int64            `json:"bound_to"`
 	OSFlavor     string            `json:"os_flavor"`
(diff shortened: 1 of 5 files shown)


apricote (approved):

jooola:
Isn't this a breaking change ? Or at least a feature?

pauhull:
The `schema` package is not guaranteed to be consistent between versions as it's intended for internal use

jooola:
Alright, do we use it externally? Or could we consider moving it into a `internal/schema` package?

pauhull:
We use them in `hcloud/schema.go` which is exported, so I don't think we can move them into an internal package

apricote:
We use it externally to implement the JSON output, as only the schema types contain the JSON tags

apricote:
I did a search for the import path, and it looks like its only used in `hcloud-go`, `cli` and the test suites for HCCM and cluster-api-provider-hetzner: https://sourcegraph.com/search?q=context:global+%22github.com/hetznercloud/hcloud-go/v2/hcloud/schema%22+-file:vendor/*+-repo:github.com/hetznercloud/hcloud-go+-repo:github.com/hetznercloud/cli&patternType=standard&sm=1&groupBy=repo

pauhull:
I made it a fix so it shows up in the changelog. I think that should be enough

jooola (approved):

PR #351, opened 2023-12-11
chore: remove go patch version

jooola:


apricote (approved):
This actually only runs `go mod tidy` and does not remove the go patch version. Go.mod has the version at 1.19 for backwards compatibility

PR #367, opened 2024-01-04
ci: add Codecov to check test coverage

pauhull:


.github/workflows/ci.yml
```diff
@@ -33,7 +33,10 @@ jobs:
       with:
         go-version: ${{ matrix.go-version }}
 
-    - run: go test -v ./...
+    - run: go test -coverprofile=coverage.txt -v -race ./...
+
+    - name: Upload coverage reports to Codecov
+      uses: codecov/codecov-action@v3
```
jooola, line 39:
```suggestion
    - codecov/codecov-action@v3
```

jooola, line 39: Does codecov understand that this is a test matrix with multiple go version ?

pauhull: Why should we remove the name? IMO it makes apparent what this step does without actually the reader actually needing to know what Codecov is

jooola (approved):

PR #371, opened 2024-01-09
feat: alias deprecated field to deprecation info struct

jooola:
Try to mitigate the removal of the deprecated field from the API, by linking the `deprecated` field value to the deprecation info object.

### DIFF
--- a/hcloud/schema/iso.go
+++ b/hcloud/schema/iso.go
@@ -1,15 +1,12 @@
 package schema
 
-import "time"
-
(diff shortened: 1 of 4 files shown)


pauhull:
Why do we keep `schema.ISO.Deprecated`?

pauhull (approved):

jooola:
We cannot back port this change because the goverter change was not implemented in v1, please ignore this warning.

PR #394, opened 2024-03-05
fix: LoadBalancerUpdateServiceOpts not converted correctly

phm07:
In the conversion from LoadBalancerUpdateServiceOpts to schema.LoadBalancerActionUpdateServiceRequest, there are conversions from slices to slice pointers. Slice pointers exist in schemas to differentiate between absent and empty.

Goverter converts nil slices to &nil. This leads to the field being marshaled to `null` in JSON (even with the `omitempty` flag) instead of being not present (See https://github.com/hetznercloud/cli/issues/702)

This PR fixes this issue by adding manual conversion methods and adds tests to account for this behavior.

### DIFF
--- a/hcloud/schema_gen.go
+++ b/hcloud/schema_gen.go
@@ -78,6 +78,8 @@ You can find a documentation of goverter here: https://goverter.jmattheis.de/
 // goverter:extend schemaFromLoadBalancerCreateOptsTargetServer
 // goverter:extend schemaFromLoadBalancerCreateOptsTargetIP
 // goverter:extend stringMapToStringMapPtr
+// goverter:extend int64SlicePtrFromCertificatePtrSlice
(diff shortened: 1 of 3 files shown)


hcloud/schema_gen.go
```diff
@@ -926,3 +928,22 @@ func mapZeroFloat32ToNil(f float32) *float32 {
 func isDeprecationNotNil(d *DeprecationInfo) bool {
 	return d != nil
 }
+
+// this is needed so that a nil slice is mapped to nil instead of &nil
```
jooola, line 932:
```suggestion
// int64SlicePtrFromCertificatePtrSlice is needed so that a nil slice is mapped to nil instead of &nil.
```

jooola (approved):

PR #396, opened 2024-03-05
fix: firewall destination IPs are omitted in schema

phm07:


jooola:
What is the rationale behind this change ?

I am not sure if this is valid. As the `*_ips` fields are only required depending on the `direction`.

phm07:
> What is the rationale behind this change ?
> 
> I am not sure if this is valid. As the `*_ips` fields are only required depending on the `direction`.

Yes, they are, but the API returns empty arrays instead of omitting the property, see here:

```json
{
  "id": 1216477,
  "name": "test",
  "labels": {},
  "created": "2024-01-23T08:56:12+00:00",
  "rules": [
    {
      "direction": "in",
      "protocol": "tcp",
      "port": "123",
      "source_ips": [
        "0.0.0.0/0",
        "::/0"
      ],
      "destination_ips": [],
      "description": null
    },
    {
      "direction": "out",
      "protocol": "tcp",
      "port": "123",
      "source_ips": [],
      "destination_ips": [
        "0.0.0.0/0",
        "::/0"
      ],
      "description": null
    }
  ],
  "applied_to": []
}
```

We should reflect this so that our marshaled schema is consistent with the API responses.

apricote (approved):

PR #397, opened 2024-03-07
feat: add volume format property

phm07:
This property was previously missing, although it was already used in VolumeCreateOpts. It's now a string pointer with common values (ext4 and xfs) available as string constants.

### DIFF
--- a/hcloud/volume.go
+++ b/hcloud/volume.go
@@ -30,7 +30,7 @@ type Volume struct {
 
 const (
 	VolumeFormatExt4 = "ext4"
-	VolumeFormatXfs  = "xfs"
+	VolumeFormatXFS  = "xfs"
 )
 
 // VolumeProtection represents the protection level of a volume.
--- a/hcloud/volume.go
+++ b/hcloud/volume.go
@@ -21,11 +21,17 @@ type Volume struct {
 	Server      *Server
 	Location    *Location
 	Size        int
+	Format      *string
 	Protection  VolumeProtection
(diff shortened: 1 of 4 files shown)


hcloud/volume.go
```diff
@@ -21,12 +21,18 @@ type Volume struct {
...
 	Protection  VolumeProtection
 	Labels      map[string]string
 	LinuxDevice string
 	Created     time.Time
 }
 
+const (
+	VolumeFormatExt4 = "ext4"
+	VolumeFormatXfs  = "xfs"
+)
```
jooola, line 34: Do you think it's worth using a custom type?

phm07: I decided against it because `VolumeCreateOpts` already used a string pointer for that property, so we would either have to make it inconsistent or a breaking change.

hcloud/volume.go
```diff
@@ -21,12 +21,18 @@ type Volume struct {
...
+	Format      *string
 	Protection  VolumeProtection
 	Labels      map[string]string
 	LinuxDevice string
 	Created     time.Time
 }
 
+const (
+	VolumeFormatExt4 = "ext4"
+	VolumeFormatXfs  = "xfs"
```
jooola, line 33:
```suggestion
	VolumeFormatXFS  = "xfs"
```

jooola (approved):

apricote:
@phm07 Can you backport this?

PR #405, opened 2024-04-17
fix: setting firewall rules not working correctly

phm07:
Because the Firewall API expects a different JSON formatting than it responds with (some fields must be omitted even though they are included in the response) we have to create an intermediate struct to marshal rules properly.

See https://github.com/hetznercloud/cli/issues/725

### DIFF
--- a/hcloud/firewall.go
+++ b/hcloud/firewall.go
@@ -298,7 +298,35 @@ type FirewallSetRulesOpts struct {
 
 // SetRules sets the rules of a Firewall.
 func (c *FirewallClient) SetRules(ctx context.Context, firewall *Firewall, opts FirewallSetRulesOpts) ([]*Action, *Response, error) {
-	reqBody := firewallSetRulesOptsToSchema(opts)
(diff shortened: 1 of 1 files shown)


hcloud/firewall.go
```diff
@@ -298,7 +298,35 @@ type FirewallSetRulesOpts struct {
 
 // SetRules sets the rules of a Firewall.
 func (c *FirewallClient) SetRules(ctx context.Context, firewall *Firewall, opts FirewallSetRulesOpts) ([]*Action, *Response, error) {
-	reqBody := firewallSetRulesOptsToSchema(opts)
+	s := firewallSetRulesOptsToSchema(opts)
+
+	// We can't use the FirewallRule struct here, because unlike in the response some fields must be omitted when empty.
+	type firewallRule struct {
```
apricote, line 304: Should we make this a `schema.FirewallRuleRequest` instead to clearly separate the API schema code from business code?

apricote (approved):

Issue #311, opened 2023-08-29
Retry transient HTTP Errors


mschfh:
Please (re)implement retries for transient HTTP error codes (502, 503, 504, etc) with sensible defaults

apricote:
Hey @mschfh,

plainly retrying requests could be a bit dangerous, as we do not have an idempotent API. Actions might be taken twice, if the actual API request arrived at the backend, but then some proxy in the middle fails the request.

Do you know of any SDKs/API Clients which do this?

alexandernst:
afaik aws's SDK does this

https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html

jooola:
Until we find out what the best approach to solve this, you may use an external library that implement a retry round tripper:

```go
package main

import (
	"net/http"

	"github.com/hashicorp/go-retryablehttp"
	"github.com/hetznercloud/hcloud-go/v2/hcloud"
)

func main() {
	token := "YOUR_API_TOKEN"

	httpClient := &http.Client{}
	httpClient.Transport = &retryablehttp.RoundTripper{}

	hcloud.NewClient(
		hcloud.WithToken(token),
		hcloud.WithHTTPClient(httpClient),
	)
}
```

PR #407, opened 2024-04-18
feat: implement actions waiter

jooola:
Implement a simpler and more versatile waiting functions for actions.

Most use cases when waiting for actions is to return early if an action fails. If all actions must be waited until completion, the users should use the `WaitForFunc` function.

If the final actions objects are needed, the users should use the `WaitForFunc` function to store the final actions using the `handleUpdate` callback.

This deprecates the `ActionClient.WatchOverallProgress`  and `ActionClient.WatchProgress` methods.

### DIFF
--- a/hcloud/action_waiter.go
+++ b/hcloud/action_waiter.go
@@ -15,8 +15,9 @@ type ActionWaiter interface {
 
 var _ ActionWaiter = (*ActionClient)(nil)
 
-// WaitForActions waits until all actions completed (succeeded or failed) by pooling the
-// API at the interval defined by [WithPollBackoffFunc].
+// WaitForFunc waits until all actions are completed by polling the API at the interval
+// defined by [WithPollBackoffFunc]. An action is considered as complete when its status is
+// either [ActionStatusSuccess] or [ActionStatusError].
 //
 // The handleUpdate callback is called every time an action is updated.
 func (c *ActionClient) WaitForFunc(ctx context.Context, handleUpdate func(update *Action) error, actions ...*Action) error {
@@ -90,13 +91,14 @@ func (c *ActionClient) WaitForFunc(ctx context.Context, handleUpdate func(update
 	return nil
 }
 
-// WaitFor waits until all actions succeeded by pooling the API at the interval
-// defined by [WithPollBackoffFunc].
+// WaitFor waits until all actions succeed by polling the API at the interval defined by
+// [WithPollBackoffFunc]. An action is considered as succeeded when its status is either
+// [ActionStatusSuccess].
 //
-// If a single action fails, the function will stop waiting and the action underlying
-// error will be returned.
+// If a single action fails, the function will stop waiting and the error set in the
+// action will be returned as an [ActionError].
 //
-// For more flexibility, see the [WaitForActionsFunc] function.
+// For more flexibility, see the [WaitForFunc] function.
 func (c *ActionClient) WaitFor(ctx context.Context, actions ...*Action) error {
 	return c.WaitForFunc(
 		ctx,
--- a/hcloud/action_waiter.go
+++ b/hcloud/action_waiter.go
@@ -1,0 +1,111 @@
+package hcloud
+
+import (
+	"context"
+	"fmt"
+	"maps"
+	"slices"
+	"time"
+)
+
+type ActionWaiter interface {
+	WaitForFunc(ctx context.Context, handleUpdate func(update *Action) error, actions ...*Action) error
+	WaitFor(ctx context.Context, actions ...*Action) error
+}
+
+var _ ActionWaiter = (*ActionClient)(nil)
+
+// WaitForActions waits until all actions completed (succeeded or failed) by pooling the
+// API at the interval defined by [WithPollBackoffFunc].
+//
+// The handleUpdate callback is called every time an action is updated.
+func (c *ActionClient) WaitForFunc(ctx context.Context, handleUpdate func(update *Action) error, actions ...*Action) error {
+	running := make(map[int64]struct{}, len(actions))
+	for _, action := range actions {
+		if action.Status == ActionStatusRunning {
+			running[action.ID] = struct{}{}
+		} else if handleUpdate != nil {
+			// We filter out already completed actions from the API polling loop; while
+			// this isn't a real update, the caller should be notified about the new
+			// state.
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/jmattheis/goverter v1.4.0
 	github.com/prometheus/client_golang v1.19.0
+	github.com/samber/lo v1.39.0
```
jooola, line 13: > Drop the extra dependency.

Yes

hcloud/action_waiter.go
```diff
@@ -0,0 +1,102 @@
...
+
+		opts := ActionListOpts{Sort: []string{"status", "id"}}
+		for actionID := range running {
+			opts.ID = append(opts.ID, actionID)
+		}
+
+		updates, err := c.AllWithOpts(ctx, opts)
+		if err != nil {
+			return nil, err
+		}
```
jooola, line 54: > The early return after AllWithOpts would cause issues if the network is wonky. Ie. if you try to create a server from snapshot on WifiOnICE
> 
> Perhaps thats also just fine and should be handled outside of this method, related https://github.com/hetznercloud/hcloud-go/issues/311

I believe the actions waiter is not responsible for handling any network flakiness or implement any retry logic. This is better handled a few layers down, probably by wrapping the http client `Do` method.

hcloud/action_waiter.go
```diff
@@ -0,0 +1,102 @@
...
+	"fmt"
+	"time"
+
+	"github.com/samber/lo"
+)
+
+type ActionWaiter interface {
+	WaitForActionsFunc(ctx context.Context, actions []*Action, handleUpdate func(update *Action)) ([]*Action, error)
+	WaitForActions(ctx context.Context, actions []*Action) ([]*Action, error)
+	WaitForAction(ctx context.Context, action *Action) (*Action, error)
```
jooola, line 15: > `actions []*Action` as argument is anoying. After creating a server I needed to contort myself to wait on all actions:
> ```
> WatchOverallProgress(ctx, append([]*Action{result.Action}, result.NextActions...)
> ```
> What do you think about accepting `actions ...*Action` instead?
> ```
> WaitForActions(ctx, result.Action, result.NextActions...)
> ```
> With the problem that you need to also do this if have just a slice (not sure we have ever encountered this in our integrations):
> ```
> WatchOverallProgress(ctx, result.NextActions)
> ```
> vs
> ```
> WaitForActions(ctx, result.NextActions...)
> ```

I already considered using variadic arguments for the actions, somehow didn't feel like it was the best API.

I think it depends on the intent behind having both `actions` and `next_actions`, if the first action fails, should we wait for the next actions ? How should the wait method handle a single action error, early exit, or wait until all actions are finished?

I believe most of the action waiting happens on a single action, so I would definitely want to simplify how we handle a single action. Having a few extra helpers for a list of actions seem sensible, but not a priority.

Using variadic arguments seem like a nice touch, I'll try to implement it, but we must answer the few questions above as well.

hcloud/action_waiter.go
```diff
@@ -0,0 +1,102 @@
...
+	result := make([]*Action, 0, len(actions))
+
+	running := make(map[int64]struct{}, len(actions))
+	for _, action := range actions {
+		if action.Status == ActionStatusRunning {
+			running[action.ID] = struct{}{}
+		} else {
+			result = append(result, action)
+		}
+	}
```
jooola, line 31: > Thinking about the Watch usecase, do you think we should emit handleUpdate calls for events which are not running at the start of the function?

Hmm, yeah, this might make sens if we think about tracking progress, but the `handleUpdate` name seem a bit confusing in that case, since the action is the same as the one passed by the user.

hcloud/action_waiter.go
```diff
@@ -0,0 +1,102 @@
...
+			opts.ID = append(opts.ID, actionID)
+		}
+
+		updates, err := c.AllWithOpts(ctx, opts)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(updates) != len(running) {
+			// Some actions do not exist in the API, fail early and prevent an infinite loop.
```
jooola, line 57: > Do you know how the actions proxy handles an unavailable backend? Does it just not return the action or is there an error in that case?

The actions proxy will return an internal server error if a service is not responding. So an action will "disappear" from the returned actions only if the ID is unknown.

go.mod
jooola: Resolved

hcloud/action_waiter.go
jooola: The new smaller API implements variadic arguments for the actions.

hcloud/action_waiter.go
jooola: Resolved

hcloud/action_waiter.go
apricote: Go actually does not allow `WaitFor(ctx, result.Action, result.NextActions...)`, and you still need to use `appends()` for this. We have decided to keep it as is to make the case of a single action or a list of actions simpler. Only a small set of endpoints returns a single action + a list of actions.

hcloud/action_waiter.go
```diff
@@ -0,0 +1,111 @@
...
+			return ctx.Err()
+		case <-time.After(c.action.client.pollBackoffFunc(retries)):
+			retries++
+		}
+
+		opts := ActionListOpts{Sort: []string{"status", "id"}}
+		for actionID := range running {
+			opts.ID = append(opts.ID, actionID)
+		}
+		slices.Sort(opts.ID)
```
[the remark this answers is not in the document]
jooola: Yes, otherwise we have some flakiness in the mocked requests URLs.

jooola:
I fixed the few comments.

I didn't use this branch in another project, I'll try it out.

apricote:
API was tested by @jooola in some of our projects:

- https://github.com/hetznercloud/csi-driver/pull/600
- https://github.com/hetznercloud/hcloud-cloud-controller-manager/pull/636
- https://github.com/hetznercloud/packer-plugin-hcloud/pull/171
- https://github.com/hetznercloud/cli/compare/main...4e73687a7867fded8a3c9c1eb809753240fce5a1

apricote (approved):
Great work @jooola! Looking forward to replace all of our home-grown implementations with this one!

PR #411, opened 2024-04-19
feat(error): show internal correlation id in error messages

apricote:
This commit changes the way error responses from the API are formatted for display to users. When available, it adds the Correlation (Trace) ID header to the error string.
    
We often have users that post the error string, but rarely do users have debug logging active. By adding the correlation ID to error messages, we can more quickly investigate why something went wrong using the Hetzner internal tracing system.
    
This is common practice in Web Apps, they usually show a request/trace ID or an encrypted blob in case of an internal server error.

### DIFF
--- a/hcloud/client.go
+++ b/hcloud/client.go
@@ -379,6 +379,10 @@ func errorFromResponse(resp *Response, body []byte) error {
 	return hcErr
 }
 
+const (
+	headerCorrelationID = "X-Correlation-Id"
+)
+
 // Response represents a response from the API. It embeds http.Response.
 type Response struct {
 	*http.Response
@@ -410,6 +414,12 @@ func (r *Response) readMeta(body []byte) error {
 	}
 
 	return nil
+}
+
+// internalCorrelationID returns the unique ID of the request as set by the API. This ID can help with support requests,
+// as it allows the people working on identify this request in particular.
+func (r *Response) internalCorrelationID() string {
+	return r.Header.Get(headerCorrelationID)
 }
 
 // Meta represents meta information included in an API response.
--- a/hcloud/error.go
+++ b/hcloud/error.go
@@ -100,6 +100,13 @@ type Error struct {
 }
 
 func (e Error) Error() string {
+	if resp := e.Response(); resp != nil {
+		correlationID := resp.internalCorrelationID()
+		if correlationID != "" {
+			// For easier debugging, the error string contains the Correlation ID of the response.
+			return fmt.Sprintf("%s (%s) (Correlation ID: %s)", e.Message, e.Code, correlationID)
+		}
+	}
 	return fmt.Sprintf("%s (%s)", e.Message, e.Code)
 }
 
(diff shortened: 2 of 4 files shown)


hcloud/client.go
```diff
@@ -410,6 +414,12 @@ func (r *Response) readMeta(body []byte) error {
 	return nil
 }
 
+// InternalCorrelationID returns the unique ID of the request as set by the API. This ID can help with support requests,
+// as it allows the people working on identify this request in particular.
+func (r *Response) InternalCorrelationID() string {
+	return r.Header.Get(headerCorrelationID)
+}
+
```
apricote, line 422: I am not sure if we should add this to our public interface. There is no guarantee that this header will always be exposed in the API. I am fine with removing this from the public interface.

hcloud/error.go
```diff
@@ -100,6 +101,15 @@ type Error struct {
 }
 
 func (e Error) Error() string {
+	if resp := e.Response(); resp != nil {
+		isServerError := resp.StatusCode >= http.StatusInternalServerError && resp.StatusCode <= 599
```
jooola, line 105:
```suggestion
		isServerError := resp.StatusCode >= 500 && resp.StatusCode <= 599
```

Maybe we should not mix the constants and the digits?

hcloud/error.go
```diff
@@ -100,6 +101,15 @@ type Error struct {
...
+		hasCorrelationID := resp.InternalCorrelationID() != ""
+		if isServerError && hasCorrelationID {
+			// For easier debugging, the error string contains the Correlation ID if the response
+			// was sent with a status code in the "server error" range (500-599).
+			return fmt.Sprintf("%s (%s) (Correlation ID: %s)", e.Message, e.Code, resp.InternalCorrelationID())
+		}
```
jooola, line 111: Why not always print the correlation ID ?

hcloud/client.go
jooola: I'd keep it private for now, until we have a better vision on how we want to do tracing in our integrations.

I am not sure if we have plans for extending the reach of our tracing capabilities, but I think it would be awesome to not stop tracing at the boundaries of the public API, but try to push it all the way in our integrations' logic.

Maybe we are allowed to generate this correlation ID, and let the API reuse it or wrap it. I'd love to have the traces that start from a hccm call and come all the way back :).

jooola (commented):
I think it's a great addition! I'll implement it in python once we agree on how we should handle this!

hcloud/error.go
apricote: If the error is probably caused by the user (4xx) there is no need to contact support and for us to check it out. Only if the server (5xx) is responsible we need to investigate further.

Of course, there are often issues where an invalid error code is sent.

apricote: Changed it to any code.

hcloud/client.go
apricote: Changed to `internalCorrelationID` (private)

jooola (approved):

PR #412, opened 2024-04-19
fix: creating firewall with rules not working correctly

phm07:
Because the Firewall API expects a different JSON formatting than it responds with (some fields must be omitted even though they are included in the response) we have to create another struct to marshal rules properly.

This should be reverted once the Firewall API has consistent response and request schemas.

See https://github.com/hetznercloud/cli/issues/725

### DIFF
--- a/hcloud/schema/firewall.go
+++ b/hcloud/schema/firewall.go
@@ -22,8 +22,8 @@ type FirewallRule struct {
 	Description    *string  `json:"description"`
 }
 
-// FirewallRuleReq defines the schema of a Firewall rule in requests.
-type FirewallRuleReq struct {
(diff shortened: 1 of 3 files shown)


hcloud/schema/firewall.go
```diff
@@ -22,6 +22,16 @@ type FirewallRule struct {
 	Description    *string  `json:"description"`
 }
 
+// FirewallRuleReq defines the schema of a Firewall rule in requests.
+type FirewallRuleReq struct {
```
apricote, line 26: Can we call this `...Request` to be sonsistent with all other structs?

apricote (dismissed):

jooola (approved):

PR #440, opened 2024-05-13
feat(exp): add `AppendNextActions` function

jooola:
This function is used to merge both a single action and the next actions returned by the API.


hcloud/exp/kit/actions/actions.go
```diff
@@ -0,0 +1,11 @@
+package actions
```
jooola, line 1: Maybe using actions as package name will conflict with a lot of variables. Maybe actionutils is better?

apricote (approved):

PR #441, opened 2024-05-13
feat(exp): add ssh key functions

jooola:
This adds a few ssh key utilities in the ssh kit package. The functions are used in multiple Hetzner Cloud projects.


hcloud/exp/kit/ssh/ssh_key.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"crypto"
+	"crypto/ed25519"
+	"encoding/pem"
+	"fmt"
+
+	"golang.org/x/crypto/ssh"
+)
+
+// GenerateKeyPair generate an new ed25519 ssh key pair, and returns the private key and
```
apricote, line 12:
```suggestion
// GenerateKeyPair generates a new ed25519 ssh key pair, and returns the private key and
```

apricote (approved):

PR #442, opened 2024-05-13
feat(exp): add hcloud client mock package

jooola:
This adds a generated mock for the `hcloudmock.Client` that will be usable for external library for testing.


jooola:
>  Although it would be nicer if we had a unified interface that the mock client would implement and also a wrapper for the regular hcloud-go client that has those methods, like we do in the hcapi2 package in the cli.

I am not sure to fully understand your suggestions, could you elaborate please?

jooola:
Implementing a mock for the client will be on hold until we can implement an interface first client. The current struct based client does not allow us to swap the client easily.

PR #452, opened 2024-06-03
feat(exp): rename `AppendNextActions` to `AppendNext`

jooola:
Use a smaller name, as we already refer to actions in the package name: 
- `actionutils.AppendNextActions ` => `actionutils.AppendNext`


apricote (approved):
Do we need to provide backwards-compatibility here? The purpose of `exp` is that we can break the API whenever we want.

jooola:
Alright, I was being conservative but I agree.

PR #459, opened 2024-06-13
feat: add truncated exponential backoff with full jitter backoff

jooola:
- Add a truncated exponential back off function with full jitter,
- The existing exponential back off function is now capped to 60s.

https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

### DIFF
--- a/hcloud/client.go
+++ b/hcloud/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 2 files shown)


jooola:
I provided a `ExponentialBackoffWithOpts` function with some knobs, but we still need to discuss the default parameters for our `ExponentialBackoff` function.

apricote (approved):
Looks good in general. Could you describe where the 24 seconds are coming from?

jooola:
This PR now primarily adds a new back off function. The change to the existing `ExponentialBackoff` function is almost negligible (cap to 60s).

The change to the default values of the existing `ExponentialBackoff` should be done in a future PR or not at all.

apricote (approved):

PR #460, opened 2024-06-18
feat(exp): add mock utils package

jooola:
Test helper to mock http request made with the hcloud-go library.


apricote (approved):
What do you think about adding a [example](https://go.dev/blog/examples) that shows how you can use this with hcloud-go?

jooola:
> What do you think about adding a [example](https://go.dev/blog/examples) that shows how you can use this with hcloud-go?

Hmm, I need an actual `*testing.T` for the handler to work. I believe I don't have access to this inside an example.

The existing test case is already an example; it will not show in the docs, though.

PR #465, opened 2024-06-19
feat(exp): add labelutils with selector

apricote:
This method is very helpful if you want to create resources and later clean them up with the same label set. Creating resources requires a `map[string]string`, but filtering resource lists takes a selector string.

### DIFF
--- a/hcloud/exp/labelutils/selector.go
+++ b/hcloud/exp/labelutils/selector.go
@@ -1,0 +1,24 @@
+package labelutils
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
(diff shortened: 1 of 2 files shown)


jooola (approved):
I was wondering, do we also support passing multiple query parameter values with single label selectors? Is that a workaround to prevent breaking our API? 

https://github.com/hetznercloud/hcloud-go/blob/0eb5e0d3d02beea6052a2800df359e6b7480e8eb/hcloud/client.go#L341

apricote:
> I was wondering, do we also support passing multiple query parameter values with single label selectors? Is that a workaround to prevent breaking our API?

Not sure, never tested that. This just builds a valid label selector string that matches our Public API.

PR #466, opened 2024-06-21
feat(exp): add envutils package

jooola:
Add a function to read value from the environment or from a file.

### DIFF
--- a/hcloud/exp/kit/envutils/env_test.go
+++ b/hcloud/exp/kit/envutils/env_test.go
@@ -9,8 +9,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func writeTmpFile(t *testing.T, tmp string, filename string, content string) string {
-	filepath := path.Join(tmp, filename)
+// nolint:unparam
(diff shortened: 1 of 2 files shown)


hcloud/exp/kit/envutils/env_test.go
```diff
@@ -0,0 +1,104 @@
...
+			},
+		},
+		{
+			name: "empty value from environment",
+			setup: func(t *testing.T, tmpDir string) {
+				t.Setenv("CONFIG", "")
+
+				// Test for precedence
+				filepath := writeTmpFile(t, tmpDir, "config", "content")
+				t.Setenv("CONFIG_FILE", filepath)
```
jooola, line 55: Do we agree that if an env var is present but empty, we do not check for the file? This feels like the less supprising.

apricote (approved):

hcloud/exp/kit/envutils/env_test.go
apricote: Agree

PR #489, opened 2024-07-19
feat(exp): add sliceutil package

jooola:
- Adds a Batches utility to split slices into batches.


apricote:
Opps. I went on a journey, and it looks like this functionality is already merged in Go stdlib and will be released with 1.23: https://go-review.googlesource.com/c/go/+/562935

jooola:
Nice, we will be to replace this utility with the std lib once we dropped go 1.22

PR #491, opened 2024-07-23
feat: allow checking multiple errors codes in `IsError`

jooola:
Useful when one need to check multiple error codes at once, without having to cast every time.

```diff
-hcloud.IsError(err, hcloud.ErrorCodeConflict) || hcloud.IsError(err, hcloud.ErrorCodeLocked)
+hcloud.IsError(err, hcloud.ErrorCodeConflict, hcloud.ErrorCodeLocked)
```

### DIFF
--- a/hcloud/error.go
+++ b/hcloud/error.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"slices"
 )
 
 // ErrorCode represents an error code returned from the API.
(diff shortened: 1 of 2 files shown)


hcloud/error.go
```diff
@@ -127,10 +128,10 @@ type ErrorDetailsInvalidInputField struct {
 }
 
 // IsError returns whether err is an API error with the given error code.
```
apricote, line 130: Can you update the docs to explain the behaviour?

> If multiple error codes are passed in, the function returns true if any of the error codes matches the passed in error.

hcloud/error.go
```diff
@@ -126,11 +127,11 @@ type ErrorDetailsInvalidInputField struct {
 	Messages []string
 }
 
-// IsError returns whether err is an API error with the given error code.
-func IsError(err error, code ErrorCode) bool {
+// IsError returns whether err is an API error with on of the given error codes.
```
jooola, line 130:
```suggestion
// IsError returns whether err is an API error with one of the given error codes.
```

apricote (approved):

PR #499, opened 2024-07-31
fix: update `NetworkSubnetType` constants

jooola:
- Update the properties documentation
- Add deprecation warning for long deprecated `NetworkSubnetTypeServer`.

### DIFF
--- a/hcloud/network.go
+++ b/hcloud/network.go
@@ -29,13 +29,13 @@ type NetworkSubnetType string
 
 // List of available network subnet types.
 const (
-	// Used to connect cloud servers.
+	// Used to connect cloud servers and load balancers.
 	NetworkSubnetTypeCloud NetworkSubnetType = "cloud"
-	// Used to connect cloud servers.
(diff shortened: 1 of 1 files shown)


hcloud/network.go
```diff
@@ -29,8 +29,15 @@ type NetworkSubnetType string
 
 // List of available network subnet types.
 const (
-	NetworkSubnetTypeCloud   NetworkSubnetType = "cloud"
-	NetworkSubnetTypeServer  NetworkSubnetType = "server"
+	// Used to connect cloud servers.
```
apricote, line 32: This subnet type also allows Cloud Load Balancers

```suggestion
	// Used to connect cloud servers and load balancers.
```

apricote (approved):

hcloud/network.go
jooola: I'll update this in the spec first.

apricote (approved):

PR #506, opened 2024-08-08
feat: allow using custom debug write functions

jooola:
Allow to configure custom functions to write the request/response debug logs.

### DIFF
--- a/hcloud/client.go
+++ b/hcloud/client.go
@@ -93,7 +93,7 @@ type Client struct {
 	applicationName         string
 	applicationVersion      string
 	userAgent               string
-	debugWriter             io.Writer
+	debugOpts               *DebugOpts
 	instrumentationRegistry prometheus.Registerer
 	handler                 handler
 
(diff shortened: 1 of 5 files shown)


jooola:
We might want to make this configurable, e.g. we have a difference between the terraform logs and our CLI.

Maybe using a logging library is easier?

hcloud/client.go
```diff
@@ -216,7 +216,28 @@ func WithApplication(name, version string) ClientOption {
...
-		client.debugWriter = debugWriter
+		client.debugOpts = defaultDebugOpts(debugWriter)
+	}
+}
+
+// DebugOpts defines the options used by [WithDebugOpts].
+type DebugOpts struct {
+	WriteRequest  func(id string, dump []byte)
+	WriteResponse func(id string, dump []byte)
+}
```
apricote, line 227: I am wondering if this is the right abstraction or if we should instead use OpenTelemetry [Logger](https://pkg.go.dev/go.opentelemetry.io/otel/log#Logger) or [`log/slog.Logger`](https://pkg.go.dev/log/slog#Logger)

jooola: I thought about using slog directly, but this gives us more flexibility. We might need to support different logging framework, or format the dump before printing it out, having those callbacks allows it.

If we use slog, we might have to use a dedicated slog Handler to be able to get each key=value of a record, and then be able to print it out. This feel overkill in comparison to the callbacks. 

I think adding open telemetry is a different topic, which I prefer not to mix with our dead simple debug output.

This PR can be backported to v1, while the open telemetry one, maybe less so.

That said, if we prefer to be conservative and not add yet another API for us to maintain over the years, I totally understand it and will try to implement open telemetry right away.

jooola: I'd prefer not have an untyped map to pass more data. If we need more flexibility in the argument list, I'd rather use a struct.

What additional data do you think we will want to pass in the future ?

jooola: We can also allow the user to hook its own handlers inside the chain, but I am afraid to give that much freedom to the users of the Client, as it will be way easier to mess with the request/response objects.

PR #521, opened 2024-08-27
feat: add per resource list actions endpoints

jooola:


jooola:
Closing, as it is not as useful as I thought. The actions returned by those endpoints are strictly scoped to the resource type.

PR #522, opened 2024-08-28
feat(exp): add `actionutil.AllForResource`

jooola:
Allow the users to fetch actions related to a specific resource.

This may be used as follows:
```go
actions, err := actionutil.AllForResource(ctx,
	client.Firewall.Action,
	hcloud.ActionListOpts{Status: []hcloud.ActionStatus{hcloud.ActionStatusRunning}},
	hcloud.ActionResourceTypeServer, server.ID,
)
if err != nil {
	return err
}
if err := client.Action.WaitFor(ctx, actions...); err != nil {
	return err
}
```

### DIFF
--- a/hcloud/exp/actionutil/actions.go
+++ b/hcloud/exp/actionutil/actions.go
@@ -1,6 +1,11 @@
 package actionutil
 
-import "github.com/hetznercloud/hcloud-go/v2/hcloud"
+import (
+	"context"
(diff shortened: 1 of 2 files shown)


jooola:
Here is an example: https://github.com/hetznercloud/packer-plugin-hcloud/compare/main...tmp-use-actionutil.allforresource

PR #530, opened 2024-09-06
ci: check for tidied go.mod

phm07:
This PR adds a CI check for `go mod tidy` as we [already do in the CLI](https://github.com/hetznercloud/cli/blob/main/.github/workflows/ci.yml#L76-L94).


jooola:
I am against enforcing this, this will lead to git conflict that will either have to be retried with renovate or resolved manually.

I prefer to be able to merge multiple dependency upgrade and tidy every once in a while.

jooola:
> Once a dependency is merged, renovate will rebase and re-run go mod tidy on the other PRs anyway.

That's one of my points, I don't want to wait for Renovate.

apricote:
I am also against this, but for another reason. If we always run `go mod tidy` we force all dependents to upgrade their dependencies as well, to the versions we have defined here.

This feels unnecessary if we do not rely on any new features of a lib (at which point go should automatically detect that it does not work and remove the old version from `go.sum`).

jooola:
I think the workflow in itself is also something we are against. Having this extra possibility of failing CI is annoying.

A solution in between could be to run a go mod tidy task every week and commit this. But I think a manual tidy can be done occasionally, when it needed.

jooola:
As discussed with @phm07, he is taking over the maintenance of the dependencies.

Since I'll not be the one merging the updates, I don't see any blocker for merging this.

jooola (approved):

PR #539, opened 2024-10-21
feat: include action ID in action error string

jooola:
This allows us to easily debug the failed action from user provided logs.


apricote (approved):
Could you add a unit test for this?

PR #552, opened 2024-11-11
fix: do not send primary IPs ID opts to the API

jooola:
The ID is not part of the request body docs, but was still send. This removes the ID from the request body.


jooola:
Should be backported.

apricote (approved):

PR #556, opened 2024-11-13
Check for nil in (c *ServerClient) ChangeAliasIPs

Difrex:
In some cases we can pass a nil opts.Network/*Server to ChangeAliasIPs func. Don't panic, return error instead.


jooola:
Hey, thanks for the contribution.

I am hesitant to merge those checks just in here while the rest of the library calls do not have them. I agree that the caller of this function doesn't know which values are required, but using the docs should help: https://docs.hetzner.cloud/#server-actions-change-alias-ips-of-a-network

PR #558, opened 2024-11-20
feat(exp): add mock requests after mock server creation

jooola:
This allows to extend the list of requests the server expects to respond to.

This also REPLACES the `mockutil.Server` function with a struct. You can create a new Server using the `mockutil.NewServer` function.


jooola:
The `mockutil.Server` function was not being used in our integrations, only the handler.

apricote (approved):

PR #559, opened 2024-11-20
chore: move tools in a module to reduce root module bloat

jooola:
This moves the tools dependencies in a dedicated module to reduce the bloat of the root module.

This should also reduce the amount of failed check generated CI jobs in renovate PRs, by not sharing the deps between our library and the tools that we are using.

The mix of make files and go generate seems unnecessarily complex, we might want to move everything to make files if this turns out to be painful.


jooola:
Yes, I agree that the simplicity of `go run` is nice.

Part of the reason why I did this is that we have those jobs failing, a bit too often, for me to tidy the files manually every time:

https://github.com/hetznercloud/hcloud-go/actions/runs/11889479028/job/33126124091?pr=550#step:4:28

apricote:
Go 1.24 will add a `tool` directive to `go.mod` to specify these external binaries: https://tip.golang.org/doc/go1.24#go-command

apricote:
Probably, not sure how Go <1.24 would handle these directives and if Go 1.24 only accepts them if the `go.mod` specifies `1.24` as the min version.

jooola:
Superseded by #847

PR #571, opened 2024-12-19
fix: allow getting resources with number as name

phm07:
If the name of a user-named resource was numeric, it was impossible to `Get` it from its name, since a number would always be interpreted as an ID. This PR fixes this behavior on user-namable resources by continuing to check if a resource with the given `idOrName` exists even though the `idOrName` is numeric and no resource with the given ID was found.

See https://github.com/hetznercloud/cli/issues/874

### DIFF
--- a/hcloud/iso.go
+++ b/hcloud/iso.go
@@ -71,7 +71,10 @@ func (c *ISOClient) GetByName(ctx context.Context, name string) (*ISO, *Response
 // Get retrieves an ISO by its ID if the input can be parsed as an integer, otherwise it retrieves an ISO by its name.
 func (c *ISOClient) Get(ctx context.Context, idOrName string) (*ISO, *Response, error) {
 	if id, err := strconv.ParseInt(idOrName, 10, 64); err == nil {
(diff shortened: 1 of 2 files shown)


hcloud/certificate.go
```diff
@@ -130,7 +130,10 @@ func (c *CertificateClient) GetByName(ctx context.Context, name string) (*Certif
 // retrieves a Certificate by its name. If the Certificate does not exist, nil is returned.
 func (c *CertificateClient) Get(ctx context.Context, idOrName string) (*Certificate, *Response, error) {
 	if id, err := strconv.ParseInt(idOrName, 10, 64); err == nil {
-		return c.GetByID(ctx, id)
+		cert, res, err := c.GetByID(ctx, id)
+		if cert != nil {
+			return cert, res, err
+		}
```
jooola, line 136: We should probably check for the error, if we have one, we might not want to spam the API.

jooola:
I think the ISOs can be named by the customer, it is missing from this PR.

jooola (approved):

PR #585, opened 2025-01-30
docs: add links to api changelog

jooola:
Users must subscribe to our changelog, or they might not be notified about additions, deprecations and removals.


jooola:
image

Issue #586, opened 2025-02-01
panic: runtime error: invalid memory address or nil pointer dereference (on missing network)


lirtistan:
### TL;DR

I try to write a FloatingIP failover for my keepalived setup and encountered an hidden error (can ONLY be seen if you turn OFF your network), this happend on my laptop today (my provider rebooted the router).

My test program.
```go
package main

import (
	"context"
	"flag"

	"ebony.local/enrico/logger"
	"github.com/hetznercloud/hcloud-go/v2/hcloud"
)

var log = logger.SimpleLogger()
var secret = "XXXXXXXXXXXXXXXXXXXXX"
var myName = ""
var floatingIP = "HA"

func main() {
	client := hcloud.NewClient(
		hcloud.WithToken(secret),
		hcloud.WithRetryOpts(hcloud.RetryOpts{
			MaxRetries: 1,
		}),
	)

	floatingIP, response, err := client.FloatingIP.GetByName(context.TODO(), "HA")
	if err != nil {
		log.Fatal(err)
	}

	if response.StatusCode == 200 {
		log.Debugf("%#v", floatingIP.Server)
	}
}
```

This is showing if the network is gone (unreachable) ...
```
[15:55:12] enrico@ebony:~/Projects/hetzner/hcloud-tools/floating-ip-failover$ go run -tags=debug .
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7885b4]

goroutine 1 [running]:
github.com/hetznercloud/hcloud-go/v2/hcloud.(*retryHandler).Do(0xc00007cf60, 0xc0001ce3c0, {0x0, 0x0})
	/home/enrico/.asdf/installs/golang/1.23.4/packages/pkg/mod/github.com/hetznercloud/hcloud-go/v2@v2.19.0/hcloud/client_handler_retry.go:46 +0x174
github.com/hetznercloud/hcloud-go/v2/hcloud.(*parseHandler).Do(0xc0000f4d00?, 0x915060?, {0x7b5160, 0xc000010918})
	/home/enrico/.asdf/installs/golang/1.23.4/packages/pkg/mod/github.com/hetznercloud/hcloud-go/v2@v2.19.0/hcloud/client_handler_parse.go:23 +0x3b
github.com/hetznercloud/hcloud-go/v2/hcloud.(*Client).Do(...)
[...]

PR #587, opened 2025-02-03
fix: ignore nil backoff function when configuring poll or retry options

jooola:
Ensure the back off function is not nil when a user only configures the `RetryOpts.MaxRetries` value.

Fixes to #586


jooola:
I am not sure if this is the best fix, should we instead not retry and do a nil check in the retry handler? This way, users may disable the feature entirely?

apricote (approved):

PR #609, opened 2025-03-11
fix: ignore empty string when setting endpoint

jooola:
Since we cannot return an error, we should never accept an empty string as a valid endpoint.

The current behavior is difficult to debug, when we set the endpoint to an empty string, the client fails with the `unsupported protocol scheme` erorr. 

This might be a bit magic, but between having a panic and ignore the input string, I prefer the latter.

This is a great candidate for a potential v3 of the client, were we can validate options.

### DIFF
--- a/hcloud/client.go
+++ b/hcloud/client.go
@@ -124,7 +124,9 @@ type ClientOption func(*Client)
 // WithEndpoint configures a Client to use the specified API endpoint.
 func WithEndpoint(endpoint string) ClientOption {
 	return func(client *Client) {
-		client.endpoint = strings.TrimRight(endpoint, "/")
+		if endpoint != "" {
+			client.endpoint = strings.TrimRight(endpoint, "/")
+		}
 	}
 }
 


hcloud/client.go
```diff
@@ -124,7 +124,9 @@ type ClientOption func(*Client)
 // WithEndpoint configures a Client to use the specified API endpoint.
```
jooola, line 124: I'll update the docs with the new behavior if we agree on it.

apricote:
> This might be a bit magic, but between having a panic and ignore the input string, I prefer the latter.

I am undecided on that. I think silently ignoring this configuration error is bad, as it might lead to a client talking to the wrong endpoint. On the other hand this is probably unused by nearly everyone outside of Hetzner Cloud, so it does not really matter much.

What do you think about a warning message to stderr in this case? This makes it at least discoverable that something unintended is happening.

jooola:
I would wait for the v3 to find the best solution and not implement a short term solution. Maybe with the event recorder we could send such thing but it is not there yet.

Closing for now;

PR #615, opened 2025-03-12
refactor: use generic request for all `GetByID` calls

jooola:
Uses the new generic requests function to reduce the code duplication for the `GetByID` calls.


jooola (commented):
Just did a review myself, and looks good.

apricote (approved):

PR #616, opened 2025-03-12
refactor: use generic request for all `List` calls

jooola:
Uses the new generic requests function to reduce the code duplication for the `List` calls.


jooola (commented):
Just did a review myself, and looks good :)

PR #619, opened 2025-03-12
refactor: use generic request for `Get` calls

jooola:
Uses the new generic requests function to reduce the code duplication for the `Get` calls.

### DIFF
--- a/hcloud/client_helper.go
+++ b/hcloud/client_helper.go
@@ -1,5 +1,10 @@
 package hcloud
 
+import (
+	"context"
+	"strconv"
+)
+
 // allFromSchemaFunc transform each item in the list using the FromSchema function, and
 // returns the result.
 func allFromSchemaFunc[T, V any](all []T, fn func(T) V) []V {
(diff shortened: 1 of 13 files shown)


hcloud/client_helper.go
```diff
@@ -51,3 +56,28 @@ func firstByName[T any](name string, listFn func() ([]*T, *Response, error)) (*T
 
 	return firstBy(listFn)
 }
+
+// getByIDOrName fetches the resource by ID when the identifier is an integer, otherwise
+// by Name. To support resources that have a integer as Name, an additional attempt is
+// made to fetch the resource by Name using the ID.
+//
+// This function is only meaningful for resources created by users.
```
apricote, line 64:
```suggestion
// This function is only meaningful for resources created by users.
```

Can you explain what you mean by this? Is this meant as a recommendation to use the `ID` whereever possible?

apricote (approved):

hcloud/client_helper.go
jooola: I rewrote the comment to make it more explicit.

PR #624, opened 2025-03-17
refactor: use generic request for actions calls

jooola:


apricote (approved):
A few calls where the wrong schema struct is being used. All of them were already wrong before this commit, I just noticed it while reviewing, not necessary to fix in this PR.

PR #626, opened 2025-03-17
feat: redefine `api_endpoint` metric labels

jooola:
- Keep path variables hints in the api_endpoint label
- Do not compute the api_endpoint label from the URL path
- Pass down the api_endpoint label using a context value


apricote (approved):
Please add a note to the changelog for this

hcloud/internal/instrumentation/metrics.go
```diff
@@ -73,8 +74,11 @@ func (i *Instrumenter) instrumentRoundTripperEndpoint(counter *prometheus.Counte
 	return func(r *http.Request) (*http.Response, error) {
 		resp, err := next.RoundTrip(r)
 		if err == nil {
-			statusCode := strconv.Itoa(resp.StatusCode)
-			counter.WithLabelValues(statusCode, strings.ToLower(resp.Request.Method), preparePathForLabel(resp.Request.URL.Path)).Inc()
+			counter.WithLabelValues(
+				strconv.Itoa(resp.StatusCode),
+				strings.ToLower(resp.Request.Method),
+				ctxutil.OpPath(r.Context()),
```
lukasmetzner, line 80: With hcloud-go v2.20, the csi-driver is now panicking upon setting up the metadata client with instrumentation and querying the hostname ([ref](https://github.com/hetznercloud/csi-driver/actions/runs/13955234652/job/39064519733?pr=912)). AFAICT it's because the metadata client does use any context and `ctx = ctxutil.SetOpPath(ctx, opPath)` is never called.

apricote: Will be fixed by #635

PR #633, opened 2025-03-19
fix: move primary ip client request/response structs to schema package

jooola:
- Move the request and response schema to the schema package
- Rename structs to follow the struct naming conventions:
  - in the schema package we hold `Request` and `Response` structs
  - in the hcloud package we hold `Opts` and `Result` structs
- Fix some response schema being used in the wrong places
- Deprecate the `Result` structs present in the hcloud package (should never be used by users)


apricote (approved):
Thanks for the additional tests.

PR #651, opened 2025-04-10
test: setup pre-commit

jooola:
- Replaces multiple hand written check with pre-commit hooks.
- Introduce the prettier pre-commit hook and format repository files.


jooola:
An alternative would be to have all those commands in a Makefile that we reuse in the CI.

lukasmetzner (approved):

PR #667, opened 2025-06-16
feat(firewall): add `applied_to_resources` property

### LINKED ISSUE #660: "applied_to_resources" is missing in FirewallResource
### TL;DR

The `applied_to_resources` is missing in the `FirewallResource` struct.

### Expected behavior

`FirewallResourceLabelSelector` contains the field `applied_to_resources` which is described in https://docs.hetzner.cloud/#firewalls-get-all-firewalls (Response Example line 35)

### Observed behavior

The field is missing

### Minimal working example

_No response_

### Log output

```plain text

```

### Additional information

_No response_

phm07:
Fixes #660

### DIFF
--- a/hcloud/firewall.go
+++ b/hcloud/firewall.go
@@ -74,21 +74,8 @@ type FirewallResource struct {
 	Server        *FirewallResourceServer
 	LabelSelector *FirewallResourceLabelSelector
 	// AppliedToResources is only used if the Type is FirewallResourceTypeLabelSelector.
-	AppliedToResources []FirewallAppliedResource
-}
-
-// FirewallAppliedResourceType specifies the type of resource that the Firewall is applied to.
-type FirewallAppliedResourceType string
-
-const (
-	// FirewallAppliedResourceTypeServer specifies a Server that a Firewall is applied to.
-	FirewallAppliedResourceTypeServer FirewallAppliedResourceType = "server"
-)
-
-// FirewallAppliedResource represents a resource that the Firewall is applied to.
-type FirewallAppliedResource struct {
-	Type   FirewallAppliedResourceType
-	Server *FirewallResourceServer
+	// FirewallResources contained here cannot be of type FirewallResourceTypeLabelSelector.
+	AppliedToResources []FirewallResource
 }
 
 // FirewallResourceServer represents a Server to apply a Firewall on.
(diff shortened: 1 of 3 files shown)


hcloud/firewall.go
```diff
@@ -70,9 +70,24 @@ const (
 
 // FirewallResource represents a resource to apply the new Firewall on.
 type FirewallResource struct {
-	Type          FirewallResourceType
-	Server        *FirewallResourceServer
-	LabelSelector *FirewallResourceLabelSelector
+	Type               FirewallResourceType
+	Server             *FirewallResourceServer
+	LabelSelector      *FirewallResourceLabelSelector
+	AppliedToResources []FirewallAppliedResource
```
apricote: Can you document that this field is only filled for `Type==label_selector`?

hcloud/firewall.go
```diff
@@ -70,9 +70,24 @@ const (
...
+}
+
+// FirewallAppliedResourceType specifies the type of resource that the Firewall is applied to.
+type FirewallAppliedResourceType string
+
+const (
+	// FirewallAppliedResourceTypeServer specifies a Server that a Firewall is applied to.
+	FirewallAppliedResourceTypeServer FirewallAppliedResourceType = "server"
+)
```
apricote: Not sure if we actually need this. Is there a reason we should not use `FirewallResourceTypeServer` instead?

phm07: We talked about it and collected the following points:
- Creating a new enum type
    - Benefits: We are safe in case in the future a new "applied to" resource is added that somehow can't be assigned regularly
    - The possible values this enum can take are immediately clear by looking at the source code and documentation
    - On the other hand it adds avoidable duplication
- Creating a new struct but reusing the enum
    - The opposite of the above, we avoid duplication but can't differentiate between FirewallResources and AppliedToResources, we would need to add extra documentation about which values are possible, but this struct is ReadOnly anyway so it might not be that big of a deal
- Reusing the entire FirewallResource struct recursively
    - The least duplication
    - We already do the same with LoadBalancerTarget
    - However, you would be able to infinitely nest Firewall resources, which does not reflect the API correctly

phm07: @jooola @lukasmetzner what are your thoughts on this? Which variant do you like the most?

jooola: I would ask the backend team to get some insight. From my understanding its a "recursive" structure so I would go with the 3rd option, but I agree that it might make things confusing so 2nd option might be better.

Option 3 could be used with some additional documentation.

phm07: I talked to the backend team, they said they have no strong preference but a recursive struct would probably be better since we already do the exact same with LoadBalancerTargets.

apricote (approved):

PR #672, opened 2025-07-08
refactor: add accept application/json header to get requests

lukasmetzner:
We have a pure JSON API and only want to accept `application/json` as a valid `Content-Type`.

### DIFF
--- a/hcloud/client_handler_debug_test.go
+++ b/hcloud/client_handler_debug_test.go
@@ -27,6 +27,7 @@ func TestDebugHandler(t *testing.T) {
 GET /v1/ HTTP/1.1
 Host: api.hetzner.cloud
 User-Agent: hcloud-go/testing
+Accept: application/json
 Authorization: REDACTED
 Accept-Encoding: gzip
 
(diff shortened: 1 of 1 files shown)


apricote (commented):
Is there a reason why you only add it for `GET` requests? I would have expected this header to be set in [`func (c *Client) NewRequest()`](https://github.com/hetznercloud/hcloud-go/blob/eb70ebca400632a5d22c35ac21be420a09c37cd8/hcloud/client.go#L300C1-L300C28), where we also set the `Content-Type` header.

lukasmetzner:
> Is there a reason why you only add it for `GET` requests? I would have expected this header to be set in [`func (c *Client) NewRequest()`](https://github.com/hetznercloud/hcloud-go/blob/eb70ebca400632a5d22c35ac21be420a09c37cd8/hcloud/client.go#L300C1-L300C28), where we also set the `Content-Type` header.

No, good point. I moved it to the `NewRequest` function.

apricote (approved):

PR #683, opened 2025-08-06
feat: add support for Hetzner API endpoint

lukasmetzner:


hcloud/client.go
```diff
@@ -23,6 +23,9 @@ import (
 // Endpoint is the base URL of the API.
```
apricote: Can we update this comment?


```suggestion
// Endpoint is the base URL of the Cloud API.
```

hcloud/client.go
```diff
@@ -272,6 +287,11 @@ func NewClient(options ...ClientOption) *Client {
 
 	client.handler = assembleHandlerChain(client)
 
+	// Shallow copy of the client and overwrite of the API endpoint.
```
apricote, line 290:
```suggestion
	// Shallow copy of the client and overwrite of the API endpoint.
	// We have two "base clients" because the endpoint is only added to the requests URL 3 layers deep, and we want to avoid passing this info through all the layers. By embedding it in the client, we can easily select which "base client" is used for each "resource client".
	// We create a shallow copy so the handler chain and prometheus registry are the same values and it is transparent to the user.
```

apricote (approved):

PR #684, opened 2025-08-06
feat: add support for Storage Boxes

### LINKED ISSUE #675: feat: support for Storage Boxes
### TL;DR

Will there be support for the api.hetzner.com API for Storageboxes? Or is there another library to handle that?

### Expected behavior

Manage Storageboxes with this library.

apricote:
We collect all changes for the Storage Box support (#675) in this PR. It will only be merged when everything is implemented through smaller pull requests targetting the `storage-boxes` branch.

Closes #675

### DIFF
--- a/hcloud/storage_box.go
+++ b/hcloud/storage_box.go
@@ -81,6 +81,10 @@ const (
 )
 
 // StorageBoxClient is a client for the Storage Box API.
+//
+// See https://docs.hetzner.cloud/reference/hetzner#storage-boxes
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 type StorageBoxClient struct {
 	client *Client
 	Action *ResourceActionClient
@@ -89,6 +93,8 @@ type StorageBoxClient struct {
 // GetByID retrieves a [StorageBox] by its ID. If the [StorageBox] does not exist, nil is returned.
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-get-a-storage-box
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) GetByID(ctx context.Context, id int64) (*StorageBox, *Response, error) {
 	const opPath = "/storage_boxes/%d"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -109,6 +115,8 @@ func (c *StorageBoxClient) GetByID(ctx context.Context, id int64) (*StorageBox, 
 // GetByName retrieves a [StorageBox] by its name. If the [StorageBox] does not exist, nil is returned.
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-list-storage-boxes
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) GetByName(ctx context.Context, name string) (*StorageBox, *Response, error) {
 	return firstByName(name, func() ([]*StorageBox, *Response, error) {
 		return c.List(ctx, StorageBoxListOpts{Name: name})
@@ -120,6 +128,8 @@ func (c *StorageBoxClient) GetByName(ctx context.Context, name string) (*Storage
 //
 // When fetching by ID, see https://docs.hetzner.cloud/reference/hetzner#storage-boxes-get-a-storage-box
 // When fetching by name, see https://docs.hetzner.cloud/reference/hetzner#storage-boxes-list-storage-boxes
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) Get(ctx context.Context, idOrName string) (*StorageBox, *Response, error) {
 	return getByIDOrName(ctx, c.GetByID, c.GetByName, idOrName)
 }
@@ -148,6 +158,8 @@ func (l StorageBoxListOpts) values() url.Values {
 // when their value corresponds to their zero value or when they are empty.
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-list-storage-boxes
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) List(ctx context.Context, opts StorageBoxListOpts) ([]*StorageBox, *Response, error) {
 	const opPath = "/storage_boxes?%s"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -165,6 +177,8 @@ func (c *StorageBoxClient) List(ctx context.Context, opts StorageBoxListOpts) ([
 // All returns all [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-list-storage-boxes
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) All(ctx context.Context) ([]*StorageBox, error) {
 	return c.AllWithOpts(ctx, StorageBoxListOpts{ListOpts: ListOpts{PerPage: 50}})
 }
@@ -172,6 +186,8 @@ func (c *StorageBoxClient) All(ctx context.Context) ([]*StorageBox, error) {
 // AllWithOpts returns all [StorageBox] with the given options.
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-list-storage-boxes
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) AllWithOpts(ctx context.Context, opts StorageBoxListOpts) ([]*StorageBox, error) {
 	return iterPages(func(page int) ([]*StorageBox, *Response, error) {
 		opts.Page = page
@@ -223,6 +239,8 @@ type StorageBoxCreateResult struct {
 // is sent to the API. They are not addressable by ID or name.
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-create-a-storage-box
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) Create(ctx context.Context, opts StorageBoxCreateOpts) (StorageBoxCreateResult, *Response, error) {
 	const opPath = "/storage_boxes"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -255,6 +273,8 @@ type StorageBoxUpdateOpts struct {
 // Update updates a [StorageBox] with the given options.
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-update-a-storage-box
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) Update(ctx context.Context, storageBox *StorageBox, opts StorageBoxUpdateOpts) (*StorageBox, *Response, error) {
 	const opPath = "/storage_boxes/%d"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -278,6 +298,8 @@ type StorageBoxDeleteResult struct {
 // Delete deletes a [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-delete-a-storage-box
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) Delete(ctx context.Context, storageBox *StorageBox) (StorageBoxDeleteResult, *Response, error) {
 	const opPath = "/storage_boxes/%d"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -315,6 +337,8 @@ func (o StorageBoxFoldersOpts) values() url.Values {
 // Folders lists folders in a [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-boxes-list-folders-of-a-storage-box
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) Folders(ctx context.Context, storageBox *StorageBox, opts StorageBoxFoldersOpts) (StorageBoxFoldersResult, *Response, error) {
 	const opPath = "/storage_boxes/%d/folders?%s"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -341,6 +365,8 @@ type StorageBoxChangeProtectionOpts struct {
 // ChangeProtection changes the protection level of a [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-change-protection
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) ChangeProtection(ctx context.Context, storageBox *StorageBox, opts StorageBoxChangeProtectionOpts) (*Action, *Response, error) {
 	const opPath = "/storage_boxes/%d/actions/change_protection"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -364,6 +390,8 @@ type StorageBoxChangeTypeOpts struct {
 // ChangeType changes the type of a [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-change-type
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) ChangeType(ctx context.Context, storageBox *StorageBox, opts StorageBoxChangeTypeOpts) (*Action, *Response, error) {
 	const opPath = "/storage_boxes/%d/actions/change_type"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -387,6 +415,8 @@ type StorageBoxResetPasswordOpts struct {
 // ResetPassword resets the password of a [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-reset-password
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) ResetPassword(ctx context.Context, storageBox *StorageBox, opts StorageBoxResetPasswordOpts) (*Action, *Response, error) {
 	const opPath = "/storage_boxes/%d/actions/reset_password"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -414,6 +444,8 @@ type StorageBoxUpdateAccessSettingsOpts struct {
 // UpdateAccessSettings updates the [StorageBoxAccessSettings] of a [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-update-access-settings
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) UpdateAccessSettings(ctx context.Context, storageBox *StorageBox, opts StorageBoxUpdateAccessSettingsOpts) (*Action, *Response, error) {
 	const opPath = "/storage_boxes/%d/actions/update_access_settings"
 	ctx = ctxutil.SetOpPath(ctx, opPath)
@@ -437,6 +469,8 @@ type StorageBoxRollbackSnapshotOpts struct {
 // RollbackSnapshot rolls back a [StorageBox] to a [StorageBoxSnapshot].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-rollback-snapshot
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) RollbackSnapshot(
 	ctx context.Context,
 	storageBox *StorageBox,
@@ -475,6 +509,8 @@ type StorageBoxEnableSnapshotPlanOpts struct {
 // EnableSnapshotPlan enables a [StorageBoxSnapshotPlan] for a [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-enable-snapshot-plan
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) EnableSnapshotPlan(
 	ctx context.Context,
 	storageBox *StorageBox,
@@ -497,6 +533,8 @@ func (c *StorageBoxClient) EnableSnapshotPlan(
 // DisableSnapshotPlan disables the [StorageBoxSnapshotPlan] for a [StorageBox].
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-actions-disable-snapshot-plan
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) DisableSnapshotPlan(
 	ctx context.Context,
 	storageBox *StorageBox,
--- a/hcloud/storage_box_snapshot.go
+++ b/hcloud/storage_box_snapshot.go
@@ -33,6 +33,8 @@ type StorageBoxSnapshotStats struct {
 // GetSnapshotByID gets a [StorageBoxSnapshot] by its ID.
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-snapshots-get-a-snapshot
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) GetSnapshotByID(ctx context.Context, storageBox *StorageBox, id int64) (*StorageBoxSnapshot, *Response, error) {
 	const optPath = "/storage_boxes/%d/snapshots/%d"
 	ctx = ctxutil.SetOpPath(ctx, optPath)
@@ -53,6 +55,8 @@ func (c *StorageBoxClient) GetSnapshotByID(ctx context.Context, storageBox *Stor
 // GetSnapshotByName gets a [StorageBoxSnapshot] by its name.
 //
 // See https://docs.hetzner.cloud/reference/hetzner#storage-box-snapshots-list-snapshots
+//
+// Experimental: [StorageBoxClient] is experimental, breaking changes may occur within minor releases.
 func (c *StorageBoxClient) GetSnapshotByName(
 	ctx context.Context,
 	storageBox *StorageBox,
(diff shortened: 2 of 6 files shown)


hcloud/storage_box.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	return ActionFromSchema(respBody.Action), resp, nil
+}
+
+type StorageBoxFoldersResult struct {
+	Folders []string
+}
+
+type StorageBoxFoldersOpts struct {
+	Path string
```
phm07, line 326: This should be a pointer since it's optional

lukasmetzner: Addressed in #698

hcloud/schema/storage_box.go
```diff
@@ -0,0 +1,105 @@
...
+// creating a storage box.
+type StorageBoxCreateResponse struct {
+	StorageBox StorageBox `json:"storage_box"`
+	Action     Action     `json:"action"`
+}
+
+// StorageBoxUpdateRequest defines the schema of the request to update a storage box.
+type StorageBoxUpdateRequest struct {
+	Name   string            `json:"name,omitempty"`
+	Labels map[string]string `json:"labels,omitempty"`
```
phm07: Should be a `*map[string]string` here, otherwise it is impossible to remove all labels.

On a related note, for some reason the API requires both properties, so you have to send a name each time as well, even if you only want to update the labels. I don't know if that is intended behavior on the API side.

lukasmetzner: But maps are reference types and initalized with `nil`. Code generation with `*map[string]string` leads to this conversion function:
```go
func stringMapToStringMapPtr(m map[string]string) *map[string]string {
	if m == nil {
		return nil
	}
	return &m
}
```

So I don't think this is necessary. Should also work with like this.

lukasmetzner: Debug msg from cli without `Labels` being a pointer:
```
--- Request:
PUT /v1/storage_boxes/REDACTED HTTP/1.1
Host: api.hetzner.com
User-Agent: hcloud-cli/1.51.0-dev hcloud-go/2.22.0
Content-Length: 26
Accept: application/json
Authorization: REDACTED
Content-Type: application/json
Accept-Encoding: gzip

{"name":"cli-test-rename"}
```

lukasmetzner: Clarified in person. Will update to `*map[string]string`

hcloud/storage_box.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+
+	result.Folders = respBody.Folders
+
+	return result, resp, nil
+}
+
+// StorageBoxChangeProtectionOpts specifies options for changing the protection level of a storage box.
+type StorageBoxChangeProtectionOpts struct {
+	Delete bool
```
phm07: This should be a pointer

lukasmetzner: https://github.com/hetznercloud/cli/pull/1143#pullrequestreview-3180660601

I will leave it like this for now, until the API topic has clarified.

hcloud/storage_box.go
```diff
@@ -0,0 +1,448 @@
...
+
+	return ActionFromSchema(respBody.Action), resp, nil
+}
+
+// StorageBoxEnableSnapshotPlanOpts specifies options for enabling a snapshot plan for a storage box.
+type StorageBoxEnableSnapshotPlanOpts struct {
+	MaxSnapshots int
+	Minute       *int // Null means every minute.
+	Hour         *int // Null means every hour.
+	DayOfWeek    *int // Null means every day. // TODO: Starts with 1 for Monday and goes to 7 for Sunday
```
phm07: Should we make this a `*time.Weekday` and do the conversion work in hcloud-go?

Pros:
- Data type is clear, standard library type can be used, no deviation from conventional (cron) format, user does not have to read documentation to understand this field

Cons:
- If we choose not to "translate" the output, it would be inconsistent. If we do, we modify the API output, which could be unexpected. JSON output in the CLI for example would be different from the HTTP response as well.

apricote: I am personally in favor of using `*time.Weekday` here and mapping it to the API format in schema conversion.

Please add a comment to the field to indicate that this differs from the API

lukasmetzner: That works for me :+1:

lukasmetzner: Implemented in #716

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,217 @@
...
+		func(ctx context.Context, name string) (*StorageBoxSnapshot, *Response, error) {
+			return c.GetSnapshotByName(ctx, storageBox, name)
+		},
+		idOrName,
+	)
+}
+
+// StorageBoxSnapshotListOpts specifies options for listing Storage Box snapshots.
+type StorageBoxSnapshotListOpts struct {
+	LabelSelector string
```
phm07, line 96: Why add `LabelSelector` here instead of embedding the generic `ListOpts` and then ignoring the pagination? In the CLI we construct a ListOpts struct and then pass it down to the "specific" ListOpts.

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,217 @@
...
+	ctx = ctxutil.SetOpPath(ctx, optPath)
+
+	reqPath := fmt.Sprintf(optPath, storageBox.ID, opts.values().Encode())
+
+	respBody, resp, err := getRequest[schema.StorageBoxSnapshotListResponse](ctx, c.client, reqPath)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return allFromSchemaFunc(respBody.Snapshots, StorageBoxSnapshotFromSchema), resp, nil
```
phm07, line 141: We usually don't return the response for list endpoints.

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	subaccount := StorageBoxSubaccountFromSchema(respBody.Subaccount)
+
+	return subaccount, resp, nil
+}
+
+// StorageBoxSubaccountListOpts represents the options for listing Storage Box subaccounts.
+type StorageBoxSubaccountListOpts struct {
+	LabelSelector string
```
phm07, line 106: Same here as with snapshots

hcloud/storage_box_snapshot.go
lukasmetzner: AFAIK we return the response in the `List` call. `All` and `AllWithOpts` calls do not return the response as they use pagination.

Even though the endpoint does not provide pagination I implemented them as usual and provided `All...()` as wrappers.

lukasmetzner:
``` go
func (c *LoadBalancerClient) List(ctx context.Context, opts LoadBalancerListOpts) ([]*LoadBalancer, *Response, error)
func (c *ServerClient) List(ctx context.Context, opts ServerListOpts) ([]*Server, *Response, error)
func (c *FirewallClient) List(ctx context.Context, opts FirewallListOpts) ([]*Firewall, *Response, error)
```

phm07: Ah, got it, thanks

hcloud/storage_box_snapshot.go
lukasmetzner: Thought it might be confusing to have the option available, but without any impact.

@jooola What do you think?

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,217 @@
...
+	storageBox *StorageBox,
+) ([]*StorageBoxSnapshot, error) {
+	opts := StorageBoxSnapshotListOpts{}
+	snapshots, _, err := c.ListSnapshots(ctx, storageBox, opts)
+	return snapshots, err
+}
+
+// StorageBoxSnapshotCreateOpts specifies options for creating a Storage Box snapshot.
+type StorageBoxSnapshotCreateOpts struct {
+	Description string
```
phm07, line 174: In the API this is optional. Should it be a pointer here?

lukasmetzner: yes, fixed

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,217 @@
...
+// StorageBoxSnapshotUpdateOpts specifies options for updating a Storage Box snapshot.
+type StorageBoxSnapshotUpdateOpts struct {
+	Description *string
+	Labels      map[string]string
+}
+
+// UpdateSnapshot updates the given snapshot of a Storage Box with the provided options.
+func (c *StorageBoxClient) UpdateSnapshot(
+	ctx context.Context,
+	storageBox *StorageBox,
```
phm07: The `StorageBoxSnapshot` struct already contains the `StorageBox`, we don't need to pass it again. Same for delete

lukasmetzner: Yes, you right. I think this might be similar for subaccounts. I change that.

lukasmetzner: Addressed in #719

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,266 @@
...
+type StorageBoxSubaccountAccessSettings struct {
+	ReachableExternally bool
+	Readonly            bool
+	SambaEnabled        bool
+	SSHEnabled          bool
+	WebDAVEnabled       bool
+}
+
+// StorageBoxSubaccountAccessSettingsOpts represents the options for access settings of a Storage Box subaccount.
+type StorageBoxSubaccountAccessSettingsOpts struct {
```
phm07: I think this struct is only used for create opts. Should we name it `StorageBoxSubaccountAccessSettingsCreateOpts` instead?

lukasmetzner: I would go for `StorageBoxSubaccountCreateOptsAccessSettings`, which is similar to other resources, e.g., `LoadBalancerCreateOptsTargetLabelSelector`.

hcloud/storage_box_snapshot.go
jooola: I think endpoints without pagination are becoming a thing, so i prefer not to embed the `ListOpts` and start supporting this use case.

hcloud/storage_box.go
```diff
@@ -0,0 +1,462 @@
...
+	"time"
+
+	"github.com/hetznercloud/hcloud-go/v2/hcloud/exp/ctxutil"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud/schema"
+)
+
+// StorageBox represents a storage box in Hetzner.
+type StorageBox struct {
+	ID             int64
+	Username       *string
```
jooola: For properties that are unknown during the initialization phase, could we use go's default empty string instead of a pointer? Has this question already been raised?

The initialization phase being a rare occurrence (only in the create response), I would argue not having a pointer makes it easier to use and we don't diverge too much from the API.

hcloud/storage_box.go
```diff
@@ -0,0 +1,462 @@
...
+// Folders lists folders in a storage box.
+func (c *StorageBoxClient) Folders(ctx context.Context, storageBox *StorageBox, opts StorageBoxFoldersOpts) (StorageBoxFoldersResult, *Response, error) {
+	const opPath = "/storage_boxes/%d/folders"
+	ctx = ctxutil.SetOpPath(ctx, opPath)
+
+	reqPath := fmt.Sprintf(opPath, storageBox.ID)
+
+	if len(opts.values()) > 0 {
+		reqPath += "?" + opts.values().Encode()
+	}
```
jooola: Seems overly complicated, we should be bale to have an empty query parameter.



```suggestion
	const opPath = "/storage_boxes/%d/folders?%s"
	ctx = ctxutil.SetOpPath(ctx, opPath)

	reqPath := fmt.Sprintf(opPath, storageBox.ID, opts.values().Encode())
```

hcloud/storage_box.go
```diff
@@ -0,0 +1,462 @@
...
+	reqBody := SchemaFromStorageBoxChangeProtectionOpts(opts)
+
+	respBody, resp, err := postRequest[schema.ActionGetResponse](ctx, c.client, reqPath, reqBody)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	action := ActionFromSchema(respBody.Action)
+
+	return action, resp, nil
```
jooola:
```suggestion
	return ActionFromSchema(respBody.Action), resp, nil
```

hcloud/storage_box.go
```diff
@@ -0,0 +1,462 @@
...
+	reqBody := SchemaFromStorageBoxChangeTypeOpts(opts)
+
+	respBody, resp, err := postRequest[schema.ActionGetResponse](ctx, c.client, reqPath, reqBody)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	action := ActionFromSchema(respBody.Action)
+
+	return action, resp, nil
```
jooola:
```suggestion
	return ActionFromSchema(respBody.Action), resp, nil
```

hcloud/storage_box.go
```diff
@@ -0,0 +1,462 @@
...
+	reqBody := SchemaFromStorageBoxResetPasswordOpts(opts)
+
+	respBody, resp, err := postRequest[schema.ActionGetResponse](ctx, c.client, reqPath, reqBody)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	action := ActionFromSchema(respBody.Action)
+
+	return action, resp, nil
```
jooola:
```suggestion
	return ActionFromSchema(respBody.Action), resp, nil
```

hcloud/storage_box.go
```diff
@@ -0,0 +1,462 @@
...
+	reqBody := SchemaFromStorageBoxUpdateAccessSettingsOpts(opts)
+
+	respBody, resp, err := postRequest[schema.ActionGetResponse](ctx, c.client, reqPath, reqBody)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	action := ActionFromSchema(respBody.Action)
+
+	return action, resp, nil
```
jooola:
```suggestion
	return ActionFromSchema(respBody.Action), resp, nil
```

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,215 @@
...
+
+	"github.com/hetznercloud/hcloud-go/v2/hcloud/exp/ctxutil"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud/schema"
+)
+
+// StorageBoxSnapshot represents a snapshot of a Storage Box.
+type StorageBoxSnapshot struct {
+	ID          int64
+	Name        string
+	Description *string
```
jooola:
```suggestion
	Description string
```

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,215 @@
...
+	storageBox *StorageBox,
+) ([]*StorageBoxSnapshot, error) {
+	opts := StorageBoxSnapshotListOpts{}
+	snapshots, _, err := c.ListSnapshots(ctx, storageBox, opts)
+	return snapshots, err
+}
+
+// StorageBoxSnapshotCreateOpts specifies options for creating a Storage Box snapshot.
+type StorageBoxSnapshotCreateOpts struct {
+	Description *string
```
jooola:
```suggestion
	Description string
```

Empty description can be omitted in the create call.

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,215 @@
...
+	reqBody := SchemaFromStorageBoxSnapshotUpdateOpts(opts)
+
+	respBody, resp, err := putRequest[schema.StorageBoxSnapshotUpdateResponse](ctx, c.client, reqPath, reqBody)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	updatedSnapshot := StorageBoxSnapshotFromSchema(respBody.Snapshot)
+
+	return updatedSnapshot, resp, nil
```
jooola:
```suggestion
	return StorageBoxSnapshotFromSchema(respBody.Snapshot), resp, nil
```

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,215 @@
...
+	const opPath = "/storage_boxes/%d/snapshots/%d"
+	ctx = ctxutil.SetOpPath(ctx, opPath)
+
+	reqPath := fmt.Sprintf(opPath, snapshot.StorageBox.ID, snapshot.ID)
+
+	respBody, resp, err := deleteRequest[schema.ActionGetResponse](ctx, c.client, reqPath)
+
+	action := ActionFromSchema(respBody.Action)
+
+	return action, resp, err
```
jooola:
```suggestion
	return ActionFromSchema(respBody.Action), resp, err
```

hcloud/storage_box_snapshot.go
```diff
@@ -0,0 +1,215 @@
...
+	updatedSnapshot := StorageBoxSnapshotFromSchema(respBody.Snapshot)
+
+	return updatedSnapshot, resp, nil
+}
+
+// DeleteSnapshot deletes the given snapshot of a Storage Box.
+func (c *StorageBoxClient) DeleteSnapshot(
+	ctx context.Context,
+	snapshot *StorageBoxSnapshot,
+) (*Action, *Response, error) {
```
jooola: Is this delete sub-resource call missing a result struct?

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		if IsError(err, ErrorCodeNotFound) {
+			return nil, resp, nil
+		}
+		return nil, resp, err
+	}
+
+	subaccount := StorageBoxSubaccountFromSchema(respBody.Subaccount)
+
+	return subaccount, resp, nil
```
jooola:
```suggestion
	return StorageBoxSubaccountFromSchema(respBody.Subaccount), resp, nil
```

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,262 @@
...
+func (c *StorageBoxClient) ListSubaccounts(
+	ctx context.Context,
+	storageBox *StorageBox,
+	opts StorageBoxSubaccountListOpts,
+) ([]*StorageBoxSubaccount, *Response, error) {
+	const opPath = "/storage_boxes/%d/subaccounts"
+	ctx = ctxutil.SetOpPath(ctx, opPath)
+
+	reqPath := fmt.Sprintf(opPath, storageBox.ID)
+	reqPath = fmt.Sprintf("%s?%s", reqPath, opts.values().Encode())
```
jooola:
```suggestion
	const opPath = "/storage_boxes/%d/subaccounts?%s"
	ctx = ctxutil.SetOpPath(ctx, opPath)

	reqPath := fmt.Sprintf(opPath, storageBox.ID, opts.values().Encode())
```

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,262 @@
...
+	subaccounts, _, err := c.ListSubaccounts(ctx, storageBox, opts)
+	return subaccounts, err
+}
+
+// StorageBoxSubaccountCreateOpts represents the options for creating a Storage Box subaccount.
+type StorageBoxSubaccountCreateOpts struct {
+	Password       string
+	HomeDirectory  *string
+	AccessSettings *StorageBoxSubaccountCreateOptsAccessSettings
+	Description    *string
```
jooola:
```suggestion
	Description    string
```

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,262 @@
...
+	reqBody := SchemaFromStorageBoxSubaccountUpdateOpts(opts)
+
+	respBody, resp, err := putRequest[schema.StorageBoxSubaccountUpdateResponse](ctx, c.client, reqPath, reqBody)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	updatedSubaccount := StorageBoxSubaccountFromSchema(respBody.Subaccount)
+
+	return updatedSubaccount, resp, nil
```
jooola:
```suggestion
	return StorageBoxSubaccountFromSchema(respBody.Subaccount), resp, nil
```

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,262 @@
...
+	reqPath := fmt.Sprintf(opPath, subaccount.StorageBox.ID, subaccount.ID)
+
+	respBody, resp, err := deleteRequest[schema.ActionGetResponse](ctx, c.client, reqPath)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	action := ActionFromSchema(respBody.Action)
+
+	return action, resp, nil
```
jooola:
```suggestion
	return ActionFromSchema(respBody.Action), resp, nil
```

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,262 @@
...
+	updatedSubaccount := StorageBoxSubaccountFromSchema(respBody.Subaccount)
+
+	return updatedSubaccount, resp, nil
+}
+
+// DeleteSubaccount deletes a subaccount from a Storage Box.
+func (c *StorageBoxClient) DeleteSubaccount(
+	ctx context.Context,
+	subaccount *StorageBoxSubaccount,
+) (*Action, *Response, error) {
```
jooola: Is this delete sub-resource call missing a result struct?

hcloud/storage_box_type.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/hetznercloud/hcloud-go/v2/hcloud/schema"
+)
+
+// StorageBoxType represents a Storage Box type in the Hetzner API.
+type StorageBoxType struct {
+	ID                     int64
+	Name                   string
+	Description            string
+	SnapshotLimit          *int
+	AutomaticSnapshotLimit *int
```
jooola, line 20: Hmm, curious why those are nullable?

hcloud/storage_box_type.go
```diff
@@ -0,0 +1,112 @@
...
+	})
+}
+
+// Get retrieves a Storage Box Type by its ID if the input can be parsed as an integer, otherwise it
+// retrieves a Storage Box Type by its name. If the Storage Box Type does not exist, nil is returned.
+func (c *StorageBoxTypeClient) Get(ctx context.Context, idOrName string) (*StorageBoxType, *Response, error) {
+	if id, err := strconv.ParseInt(idOrName, 10, 64); err == nil {
+		return c.GetByID(ctx, id)
+	}
+	return c.GetByName(ctx, idOrName)
```
jooola: Should we reuse the getByIdOrName helper?

jooola (commented):
I'll do another pass when the additional changes from #745 are merged.

hcloud/storage_box_type.go
lukasmetzner: Absolutely :+1:

hcloud/storage_box_subaccount.go
lukasmetzner: Hmm, what would we return other than the action? Or just as a general convention in hcloud-go? Neither storage boxes, nor snapshot have this struct right now. Would need change there as well.

hcloud/storage_box.go
lukasmetzner: Done

hcloud/storage_box_subaccount.go
lukasmetzner: See: https://github.com/hetznercloud/hcloud-go/pull/684#discussion_r2436432527

hcloud/storage_box.go
lukasmetzner: Sounds reasonable to me in this case.

lukasmetzner: Implemented in [8547e97](https://github.com/hetznercloud/hcloud-go/pull/684/commits/8547e9709ae3c6016af87469115a5aae3a80c23f)

Opposing opinions @apricote @phm07 ?

hcloud/storage_box_subaccount.go
jooola: With zones we decided to always return a result struct for resource and sub-resource, to make it future proof in case a new field is added.

So unless there is a strong argument against using a result struct, I would prefer using one.

hcloud/storage_box.go
jooola: I added my comment on the username only, but I believe other properties follow the same pattern (System and Server). Might be a good idea to evaluate each property.

jooola:
:information_source:  I am currently fixing a bunch of things here.

hcloud/storage_box.go
jooola: Hmm, I believe you updated the schema, not the public facing struct.

hcloud/storage_box_subaccount.go
jooola: This will be resolved with #752

hcloud/storage_box_snapshot.go
jooola: This will be resolved with #752

hcloud/storage_box.go
jooola: This will be resolved with #752

apricote: Using an empty string for all three properties works for me.

hcloud/storage_box.go
```diff
@@ -0,0 +1,463 @@
...
+)
+
+// StorageBoxClient is a client for the Storage Box API.
+type StorageBoxClient struct {
+	client *Client
+	Action *ResourceActionClient
+}
+
+// GetByID retrieves a [StorageBox] by its ID. If the [StorageBox] does not exist, nil is returned.
+func (c *StorageBoxClient) GetByID(ctx context.Context, id int64) (*StorageBox, *Response, error) {
```
jooola, line 98: - Should we include links to the upstream documentation for each client methods?
- Are we releasing this as experimental or not?

lukasmetzner: For the first point, yes thats a good idea. I will do so.

hcloud/storage_box.go
```diff
@@ -0,0 +1,463 @@
...
+type StorageBoxCreateOpts struct {
+	Name           string
+	StorageBoxType *StorageBoxType
+	Location       *Location
+	Labels         map[string]string
+	Password       string
+	// TODO: API only accepts literal public keys
+	// - should we accept our structs here and read the public key?
+	// - What to do when only the ID is set and not the public key?
+	// - Should we tell users that this is not required and they can just create an object with the public key field set?
```
lukasmetzner: There is one last todo in the code. Shall we use our SSHKey struct here? Would allow users to fetch existing SSH-Keys and use their public key field. Would also be simpler to extend in the future, if the API supports different options.

hcloud/storage_box.go
lukasmetzner: On the second point, we could mark it as Experimental for a month and move it to stable then.

lukasmetzner: The API is already released for some time now. This might not hinder adoption as much, as when releasing it in parallel.

jooola: Ok, then lets mark everything as experimental for a small period.

hcloud/storage_box.go
jooola: - I would prefer not to magically fetch the SSH Key using its ID to populate the public keys, this would be unexpected.
- Will the API eventually support passing SSH Keys by IDs? This might help us decide if we should use the existing `hcloud.SSHKey` struct.
- We could reuse the `hcloud.SSHKey` struct by providing example of how to use it, and return `argumentError` if the provided struct is invalid?

lukasmetzner: Fixed in 5ceecad

hcloud/storage_box.go
lukasmetzner: Done

Issue #685, opened 2025-08-06
feat!: update "Pricings" field to "Prices"


apricote:
### TL;DR

The naming is inconsistent between `hcloud-go` and the API.

Affects these resources:

- Server Types
- Load Balancer Types
- Storage Box Types

This is a breaking change.

### Expected behavior

.

apricote:
We could add both fields, but I feel that this confuses users even more than a single field with a slightly wrong name.

PR #687, opened 2025-08-06
feat: add support for Storage Box Types

lukasmetzner:
Reference to API docs: https://docs.hetzner.cloud/reference/hetzner#storage-box-types

### DIFF
--- a/hcloud/schema/storage_box_type.go
+++ b/hcloud/schema/storage_box_type.go
@@ -1,0 +1,33 @@
+package schema
+
+// StorageBoxType represents a Storage Box type as returned by the Hetzner Cloud API.
+type StorageBoxType struct {
+	ID                     int64                         `json:"id"`
+	Name                   string                        `json:"name"`
+	Description            string                        `json:"description"`
+	SnapshotLimit          *int                          `json:"snapshot_limit"`
+	AutomaticSnapshotLimit *int                          `json:"automatic_snapshot_limit"`
+	SubaccountsLimit       int                           `json:"subaccounts_limit"`
+	Size                   int                           `json:"size"`
+	Prices                 []PricingStorageBoxTypePrices `json:"prices"`
+	DeprecatableResource
+}
+
+// PricePerLocation represents pricing for a Storage Box type in a specific location.
+type PricingStorageBoxTypePrices struct {
+	Location     string `json:"location"`
+	PriceHourly  Price  `json:"price_hourly"`
+	PriceMonthly Price  `json:"price_monthly"`
+	SetupFee     Price  `json:"setup_fee"`
+}
+
+// StorageBoxTypeListResponse represents the response for listing Storage Box Types.
+type StorageBoxTypeListResponse struct {
+	StorageBoxTypes []StorageBoxType `json:"storage_box_types"`
+	Meta            Meta             `json:"meta"`
+}
+
+// StorageBoxTypeGetResponse represents the response for getting a single Storage Box Type.
+type StorageBoxTypeGetResponse struct {
+	StorageBoxType StorageBoxType `json:"storage_box_type"`
+}
--- a/hcloud/storage_box_type.go
+++ b/hcloud/storage_box_type.go
@@ -1,0 +1,95 @@
+package hcloud
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/hetznercloud/hcloud-go/v2/hcloud/exp/ctxutil"
+	"github.com/hetznercloud/hcloud-go/v2/hcloud/schema"
+)
+
+// StorageBoxType represents a Storage Box type in the hcloud package.
+type StorageBoxType struct {
+	ID                     int64
+	Name                   string
+	Description            string
+	SnapshotLimit          *int
+	AutomaticSnapshotLimit *int
+	SubaccountsLimit       int
+	Size                   int
+	Pricings               []StorageBoxTypeLocationPricing
+	DeprecatableResource
+}
+
+type StorageBoxTypeLocationPricing struct {
+	Location     string
+	PriceHourly  Price
+	PriceMonthly Price
+	SetupFee     Price
+}
+
+// StorageBoxTypeClient provides access to Storage Box Types in the Hetzner Cloud API.
+type StorageBoxTypeClient struct {
+	client *Client
+}
+
+// StorageBoxTypeListOpts specifies options for listing storage box types.
+type StorageBoxTypeListOpts struct {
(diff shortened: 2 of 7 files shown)


hcloud/schema/storage_box_type.go
```diff
@@ -0,0 +1,33 @@
...
+// StorageBoxType represents a Storage Box type as returned by the Hetzner Cloud API.
+type StorageBoxType struct {
+	ID                     int64                         `json:"id"`
+	Name                   string                        `json:"name"`
+	Description            string                        `json:"description"`
+	SnapshotLimit          *int                          `json:"snapshot_limit"`
+	AutomaticSnapshotLimit *int                          `json:"automatic_snapshot_limit"`
+	SubaccountsLimit       int                           `json:"subaccounts_limit"`
+	Size                   int                           `json:"size"`
+	Prices                 []PricingStorageBoxTypePrices `json:"prices"`
```
jooola:
```suggestion
	Prices                 []StorageBoxTypePrice `json:"prices"`
```

hcloud/schema/storage_box_type.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// StorageBoxTypeListResponse represents the response for listing Storage Box Types.
+type StorageBoxTypeListResponse struct {
+	StorageBoxTypes []StorageBoxType `json:"storage_box_types"`
+	Meta            Meta             `json:"meta"`
+}
+
+// StorageBoxTypeGetResponse represents the response for getting a single Storage Box Type.
+type StorageBoxTypeGetResponse struct {
```
jooola, line 30:
```suggestion
type StorageBoxTypeResponse struct {
```

hcloud/storage_box_type.go
```diff
@@ -0,0 +1,95 @@
...
+// StorageBoxType represents a Storage Box type in the hcloud package.
+type StorageBoxType struct {
+	ID                     int64
+	Name                   string
+	Description            string
+	SnapshotLimit          *int
+	AutomaticSnapshotLimit *int
+	SubaccountsLimit       int
+	Size                   int
+	Pricings               []StorageBoxTypeLocationPricing
```
jooola, line 22:
```suggestion
	Pricings               []StorageBoxTypePricing
```

Why not fix the name `Pricings`=>`Prices` before we release a to be deprecated field?

hcloud/schema/storage_box_type.go
```diff
@@ -0,0 +1,33 @@
...
+
+// StorageBoxType represents a Storage Box type as returned by the Hetzner Cloud API.
+type StorageBoxType struct {
+	ID                     int64                         `json:"id"`
+	Name                   string                        `json:"name"`
+	Description            string                        `json:"description"`
+	SnapshotLimit          *int                          `json:"snapshot_limit"`
+	AutomaticSnapshotLimit *int                          `json:"automatic_snapshot_limit"`
+	SubaccountsLimit       int                           `json:"subaccounts_limit"`
+	Size                   int                           `json:"size"`
```
apricote: Must be `int64` as 20 TB in bytes is larger than `int32` can represent.

hcloud/schema/storage_box_type.go
```diff
@@ -0,0 +1,33 @@
...
+	Location     string `json:"location"`
+	PriceHourly  Price  `json:"price_hourly"`
+	PriceMonthly Price  `json:"price_monthly"`
+	SetupFee     Price  `json:"setup_fee"`
+}
+
+// StorageBoxTypeListResponse represents the response for listing Storage Box Types.
+type StorageBoxTypeListResponse struct {
+	StorageBoxTypes []StorageBoxType `json:"storage_box_types"`
+	Meta            Meta             `json:"meta"`
```
apricote: Drop

hcloud/schema/storage_box_type.go
lukasmetzner: For other types such as Load Balancer and Servers we use the same naming scheme (except for the trailing `s`). Do we want to break here? I would also be in favor of removing the `Pricing` prefix.

- PricingLoadBalancerTypePrice
- PricingServerTypePrice

hcloud/schema/storage_box_type.go
lukasmetzner: AFAIK we use the `***GetResponse` everywhere else. Why break here?

hcloud/schema/storage_box_type.go
jooola: I'd argue this is in the schema package, we could fix them all to not have the Pricing prefix.

If we want to keep the prefix, I'd prefer not having a `s` for a list item.

hcloud/storage_box_type.go
jooola: https://github.com/hetznercloud/hcloud-go/issues/685

hcloud/schema/storage_box_type.go
jooola: My bad, mixing thing with our schema component that do not have the `Get` part, since we reuse them for the return response of the update calls for example.

Ignore my comment.

hcloud/schema/storage_box_type.go
lukasmetzner: Fine for me.

Probably miscommunication, I also prefer not having the `s` suffix.

hcloud/storage_box_type.go
lukasmetzner: Works for me.

jooola: I'd like to wait for @apricote's answer on whether we should already deprecate now or wait for the major release. Then we can decide here.

lukasmetzner: @jooola I noticed though, that you also proposed to remove the infix `Location`. Should we add this to the ticket and do it for the other fields, `ServerTypeLocationPricing` and `LoadBalancerLocationPricing`?

jooola: Hmm, indeed, we should probably keep it.

lukasmetzner: > I'd like to wait for @apricote's answer on whether we should already deprecate now or wait for the major release. Then we can decide here.

Oh, I missed your comment due to us sending at the same time. I pushed the rename now, but yeah lets wait for an additional answer from @apricote, then we can keep or revert the change.

jooola: Lets move on and keep it consistent for now.

Once #685 is resolved we can update the `storage-boxes` branch as well.

apricote (approved):

PR #694, opened 2025-08-18
feat: Storage Box CRUD operations

lukasmetzner:
Adds the new Storage Box client with the necessary types and CRUD operations.

API Reference: https://docs.hetzner.cloud/reference/hetzner#storage-boxes

### DIFF
--- a/hcloud/zz_schema.go
+++ b/hcloud/zz_schema.go
@@ -1127,22 +1127,20 @@ func (c *converterImpl) SchemaFromServerType(source *ServerType) schema.ServerTy
 func (c *converterImpl) SchemaFromStorageBox(source *StorageBox) schema.StorageBox {
 	var schemaStorageBox schema.StorageBox
 	if source != nil {
-		var schemaStorageBox2 schema.StorageBox
-		schemaStorageBox2.ID = (*source).ID
-		schemaStorageBox2.Username = (*source).Username
-		schemaStorageBox2.Status = string((*source).Status)
-		schemaStorageBox2.Name = (*source).Name
-		schemaStorageBox2.StorageBoxType = c.SchemaFromStorageBoxType((*source).StorageBoxType)
-		schemaStorageBox2.Location = c.SchemaFromLocation((*source).Location)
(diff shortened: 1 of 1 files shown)


hcloud/storage_box.go
```diff
@@ -0,0 +1,281 @@
...
+func (o StorageBoxFoldersOpts) values() url.Values {
+	vals := url.Values{}
+	if o.Path != "" {
+		vals.Add("path", o.Path)
+	}
+	return vals
+}
+
+// Lists folders in a storage box.
+func (c *StorageBoxClient) Folders(ctx context.Context, storageBox *StorageBox, opts StorageBoxFoldersOpts) (StorageBoxFoldersResult, *Response, error) {
```
lukasmetzner, line 260: I don't know if we want to name this `Folders`,  `ListFolders`,  `Content` or  `ListContent`? As files are not included, `*Folders` might be simpler to understand. I am fine with all options.

apricote: I would stay with the name of the operation in the API: "List content of Storage Box" ->


```go
func (c *StorageBoxClient) ListContent() *StorageBoxListContentResponse

type StorageBoxListContentResponse struct {
  Folders []string
}
```

jooola: I'd rather look at the operation id :), maybe the backend team has a better idea of how this might evolve.

We either need to fix the spec or update the function name.

phm07: I would personally expect a function called `ListContent` to also return file names, not only directories. Also, is it listing contents recursively?

jooola: I'd also not include any `Get`, `List` prefixes.

lukasmetzner: @phm07 No it is not recursive. It accepts an optional query parameter for selecting a subpath.

lukasmetzner: @jooola With operation ID you mean `folders`?

Agreeing with @phm07, that I'd expect files when calling a method called  `Content`.

PR #698, opened 2025-08-20
fix: path in folders opts should be optional

lukasmetzner:


hcloud/storage_box.go
```diff
@@ -245,13 +245,13 @@ type StorageBoxFoldersResult struct {
 }
 
 type StorageBoxFoldersOpts struct {
-	Path string
+	Path *string
```
apricote, line 248: We are not usually handling optional fields in `Opts` with a pointer. For comparison take a look at [ServerListOpts](https://github.com/hetznercloud/hcloud-go/blob/475a81a94734adfeb49e24a6d794117d264052db/hcloud/server.go#L222-L242).

As long as the empty string has no useful meaning in the API that is different from not setting the value, its fine to use a non-pointer and `o.Path != ""` here.

lukasmetzner: Alright :+1:  I guess than we can close this PR.

PR #701, opened 2025-08-22
feat: Storage Box actions

lukasmetzner:


hcloud/schema/storage_box.go
```diff
@@ -103,3 +103,35 @@ type StorageBoxUpdateResponse struct {
 type StorageBoxFoldersResponse struct {
 	Folders []string `json:"folders"`
 }
+
+type StorageBoxChangeProtectionRequest struct {
+	Delete bool `json:"delete"`
+}
+
+type StorageBoxChangeTypeRequest struct {
+	StorageBoxType int64 `json:"storage_box_type"`
```
apricote: API also supports a `name` string in this field, can you check how we have implemented this in similar situations?

lukasmetzner: Fixed in 5a4aa8db855e33bf7c01aab02a9883a58803e536 by adjusting the schema. The rest is handled via goverter.

apricote (approved):

PR #704, opened 2025-08-25
feat: Storage Box subaccounts

lukasmetzner:
Adds support for Storage Box subaccounts.

API Reference: https://docs.hetzner.cloud/reference/hetzner#subaccounts

### DIFF
--- a/hcloud/schema/storage_box_subaccount.go
+++ b/hcloud/schema/storage_box_subaccount.go
@@ -4,24 +4,24 @@ import "time"
 
 // StorageBoxSubaccount defines the schema of a Storage Box subaccount.
 type StorageBoxSubaccount struct {
-	ID             int64                               `json:"id"`
-	Username       string                              `json:"username"`
-	HomeDirectory  string                              `json:"home_directory"`
-	Server         string                              `json:"server"`
-	AccessSettings *StorageBoxSubaccountAccessSettings `json:"access_settings"`
-	Description    string                              `json:"description"`
-	Labels         map[string]string                   `json:"labels"`
-	Created        time.Time                           `json:"created"`
-	StorageBox     int64                               `json:"storage_box"`
+	ID             int64                              `json:"id"`
+	Username       string                             `json:"username"`
+	HomeDirectory  string                             `json:"home_directory"`
+	Server         string                             `json:"server"`
+	AccessSettings StorageBoxSubaccountAccessSettings `json:"access_settings"`
+	Description    string                             `json:"description"`
+	Labels         map[string]string                  `json:"labels"`
+	Created        time.Time                          `json:"created"`
+	StorageBox     int64                              `json:"storage_box"`
 }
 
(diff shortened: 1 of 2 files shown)


hcloud/schema/storage_box_subaccount.go
```diff
@@ -0,0 +1,64 @@
...
+	StorageBox     int64                               `json:"storage_box"`
+}
+
+// StorageBoxSubaccountAccessSettings defines the schema of a Storage Box subaccount's access settings.
+type StorageBoxSubaccountAccessSettings struct {
+	ReachableExternally *bool `json:"reachable_externally,omitempty"`
+	Readonly            *bool `json:"readonly,omitempty"`
+	SambaEnabled        *bool `json:"samba_enabled,omitempty"`
+	SSHEnabled          *bool `json:"ssh_enabled,omitempty"`
+	WebDAVEnabled       *bool `json:"webdav_enabled,omitempty"`
```
apricote: These are all marked as required in the spec, so can be `bool` instead of `*bool`

hcloud/schema/storage_box_subaccount.go
```diff
@@ -0,0 +1,64 @@
...
+
+import "time"
+
+// StorageBoxSubaccount defines the schema of a Storage Box subaccount.
+type StorageBoxSubaccount struct {
+	ID             int64                               `json:"id"`
+	Username       string                              `json:"username"`
+	HomeDirectory  string                              `json:"home_directory"`
+	Server         string                              `json:"server"`
+	AccessSettings *StorageBoxSubaccountAccessSettings `json:"access_settings"`
```
apricote: For `StorageBox` this is not a pointer and marked as `required`. Can you clarify with the team if this field could also be marked as `required`?

hcloud/storage_box_subaccount.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	result.Action = ActionFromSchema(respBody.Action)
+
+	result.Subaccount = &StorageBoxSubaccount{
+		ID: respBody.Subaccount.ID,
+		StorageBox: &StorageBox{
+			ID: respBody.Subaccount.StorageBox,
+		},
+	}
```
apricote: Why can you not use `StorageBoxSubaccountFromSchema()` here?

apricote (changes requested):

hcloud/storage_box_subaccount.go
lukasmetzner: You are right, implemented it using schema gen.

hcloud/schema/storage_box_subaccount.go
lukasmetzner: I will ask them :+1:

lukasmetzner: That is a mistake. A fix is on the way. I already pushed the changes.

PR #716, opened 2025-09-08
feat: map Hetzner API's day_of_week to Go's time.Weekday

lukasmetzner:


hcloud/schema_gen.go
```diff
@@ -1045,3 +1049,27 @@ func mapStorageBoxSnapshotPtrStorageBoxSnapshotID(snapshot *StorageBoxSnapshot)
 
 	return snapshot.ID
 }
+
+func mapWeekdayPtrToIntPtr(w *time.Weekday) *int {
```
jooola: Could we make those functions "storage boxes" specific please?

lukasmetzner: Yes :+1:

PR #738, opened 2025-10-01
feat: support for go1.25 and drop go1.23

jooola:


go.mod
```diff
@@ -1,15 +1,9 @@
 module github.com/hetznercloud/hcloud-go/v2
 
-// This is being kept at a lower value on purpose as raising this would require
-// all dependends to update to the new version.
-// As long as we do not depend on any newer language feature this can be kept at the current value.
-// It should never be higher than the lowest currently supported version of Go.
 // Since golang.org/x dependencies always requires version 1.(N-1), this is effectively
 // the same version we will be using. (See http://go.dev/issue/69095)
-go 1.23.0
+go 1.24.0
```
lukasmetzner, line 5: Shall we move to `1.24.7`? Or in a separate PR?

jooola: I think we should stick to a .0 release, and follow what golang is doing for their own dependencies.

https://github.com/hetznercloud/hcloud-go/pull/721/commits/7671e218e3f079a2c98ba4230d268da32ea234ac

apricote: We do not strictly need to require users to bump to `1.24.7` to use our library.

PR #745, opened 2025-10-09
feat: storage box api changes

lukasmetzner:


hcloud/schema/storage_box.go
```diff
@@ -83,39 +83,46 @@ type StorageBoxCreateRequestAccessSettings struct {
...
 type StorageBoxCreateResponse struct {
 	StorageBox StorageBox `json:"storage_box"`
 	Action     Action     `json:"action"`
 }
 
-// StorageBoxUpdateRequest defines the schema of the request to update a storage box.
+// StorageBoxUpdateRequest defines the schema of the request to update a Storage Box.
 type StorageBoxUpdateRequest struct {
-	Name   *string            `json:"name,omitempty"`
+	Name   string             `json:"name,omitempty"`
```
apricote, line 94: This change feels backward. The fields in update requests will become optional for all three resources.

lukasmetzner: The pointer was orignally there, because the API allowed empty strings to be set as a Storage Box name. This won't be the case in the future and an empty name will be ignored.

I thought this makes it somewhat easier to the user, as a pointer is not required.

We have a mixed pattern in the codebase, so I was unsure what the best option is. Both are fine for me.

```go
// StorageBoxUpdateRequest defines the schema of the request to update a Storage Box.
type StorageBoxUpdateRequest struct {
	Name   string             `json:"name,omitempty"`
        // ...
}

// NetworkUpdateRequest defines the schema of the request to update a network.
type NetworkUpdateRequest struct {
	Name                  string             `json:"name,omitempty"`
        // ...
}

type LoadBalancerUpdateRequest struct {
	Name   *string            `json:"name,omitempty"`
        // ...
}
```

hcloud/storage_box_subaccount_test.go
```diff
@@ -434,3 +427,40 @@ func TestStorageBoxSubbacountUpdateAccessSettings(t *testing.T) {
...
+		{
+			Method: "POST", Path: "/storage_boxes/42/subaccounts/13/actions/change_home_directory",
+			Status: 201,
+			Want: func(t *testing.T, r *http.Request) {
+				body, err := io.ReadAll(r.Body)
+				require.NoError(t, err)
+
+				expected := `{
+					"home_directory": "/foobar"
+				}`
```
jooola, line 444: 🤩

jooola (approved):
The changes look good, I have comments on the code that didn't change though.

hcloud/schema/storage_box.go
jooola: I prefer without pointer, reduces the complexity.

lukasmetzner: @apricote I would go ahead and merge this now, to proceed with the other review points in the main PR. We can discuss this topic there again, if you want.

PR #746, opened 2025-10-13
refactor(storage-box): simplify tests

lukasmetzner:
- remove redundant parsing of full storage box properties
- make tests more coherent between storage box and the subressources snapshot and subaccount

### DIFF
--- a/hcloud/storage_box_snapshot_test.go
+++ b/hcloud/storage_box_snapshot_test.go
@@ -65,7 +65,7 @@ func TestStorageBoxClientGetSnapshot(t *testing.T) {
 				Method: "GET", Path: "/storage_boxes/42/snapshots?name=my-resource",
 				Status: 200,
 				JSONRaw: `{
-				"snapshots": [{ "id": 42, "name": "my-resource" }]
+				"snapshots": [{ "id": 42 }]
 			}`,
 			},
 		})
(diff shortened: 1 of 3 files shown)


hcloud/storage_box_snapshot_test.go
```diff
@@ -176,8 +145,7 @@ func TestStorageBoxClientListSnapshot(t *testing.T) {
 						"is_automatic": true,
 						"labels": {
 							"environment": "prod"
-						},
-						"storage_box": 42
```
jooola, line 180: This should probably stay, no?

lukasmetzner: I wonder if we could not remove most of this, as this is testing the same deserialization as in `GetSnapshot (ByID)`. Felt a bit wrong to not include the properties I am querying for.

lukasmetzner: Sorry, I think my last comment was not so clear. With `most of this` I meant, making it look like this:

```go
JSONRaw: `{
	"snapshots": [{
		"id": 42,
		"name": "my-resource"
	}]
}`,
```

`is_automatic` and `labels` are already tested in `GetSnapshot (ByID)`.

jooola: I would remove everything that is not used, it was unclear to me why remove the storage box id while keeping the labels.

PR #750, opened 2025-10-15
feat: retrieve a server or load balancer network attachment

jooola:
Adds the `(&Server{}).PrivateNetFor(&Network{})` and `(&LoadBalancer{}).PrivateNetFor(&Network{})` methods to retrieve the private network attachment information.

### DIFF
--- a/hcloud/load_balancer.go
+++ b/hcloud/load_balancer.go
@@ -229,8 +229,8 @@ func (lb *LoadBalancer) GetDNSPtrForIP(ip net.IP) (string, error) {
 	return "", DNSNotFoundError{ip}
 }
 
-// PrivateNetFor returns the load balancer's network assignation information in the given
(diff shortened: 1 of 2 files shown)


hcloud/load_balancer.go
```diff
@@ -228,6 +229,18 @@ func (lb *LoadBalancer) GetDNSPtrForIP(ip net.IP) (string, error) {
 	return "", DNSNotFoundError{ip}
 }
 
+// PrivateNetFor returns the load balancer's network assignation information in the given
+// Network, and nil if no assignation was found.
```
apricote: This is usually called `attach`/`attachments` in the API, not `assign`/`assignation`.

jooola: Indeed, will revert to attach/attachment

PR #754, opened 2025-10-20
feat(storage-box): use SSH key struct in create opts

lukasmetzner:
`ExampleStorageBoxClient_Create` and `ExampleStorageBoxClient_Create` are two examples, which can be expanded in the `func (*StorageBoxClient) [Create]` section of the https://pkg.go.dev/ docs.

### DIFF
--- a/hcloud/schema_gen.go
+++ b/hcloud/schema_gen.go
@@ -401,7 +401,7 @@ type converter interface {
 	// goverter:map DayOfWeek | mapStorageBoxIntPtrToWeekdayPtr
 	StorageBoxSnapshotPlanFromSchema(schema.StorageBoxSnapshotPlan) StorageBoxSnapshotPlan
 	SchemaFromStorageBox(*StorageBox) schema.StorageBox
-	// goverter:map SSHKeys | mapSSHKeyPtrSliceToPublicKeyString
+	// goverter:map SSHKeys | mapSSHKeyPtrSliceToPublicKeySlice
 	SchemaFromStorageBoxCreateOpts(StorageBoxCreateOpts) schema.StorageBoxCreateRequest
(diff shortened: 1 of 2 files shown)


hcloud/schema_gen.go
```diff
@@ -1087,6 +1088,18 @@ func locationFromServerTypeLocationSchema(serverTypeLocation schema.ServerTypeLo
...
+	publicKeys := make([]string, 0, len(s))
+	for _, key := range s {
+		if key == nil {
+			continue
+		}
+		publicKeys = append(publicKeys, key.PublicKey)
+	}
+
+	return publicKeys
+}
```
jooola: Does goverter understand when only mapping one to one instead of many to many ?

hcloud/storage_box.go
```diff
@@ -171,14 +172,28 @@ type StorageBoxCreateOpts struct {
...
+		}
+		if key.PublicKey == "" {
+			errs = append(errs, missingField(key, "PublicKey"))
+		}
+	}
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+	return nil
+}
```
jooola, line 187: I think we usually return early and do not collect multiple errors

hcloud/schema_gen.go
lukasmetzner: No, I could not make it work. It should be possible with a general `extend` conversion setting, but this would then always use `SSHKey.PublicKey` for any `*SSHKey` to `string` conversion.

lukasmetzner: https://goverter.jmattheis.de/guide/configure-nested#slices-and-maps

Does also not seem to work when converting to a basic type.

jooola: Check

PR #760, opened 2025-10-31
ci: add link check workflow

### LINKED ISSUE #755: Link Validator in CI
### TL;DR

Add `lychee` to validate that all links are working.

### Expected behavior

We have started to include links to the API docs in new API client methods. We should add a validator to make sure these links are actually valid.

I found that `lychee` works well and can even check fragments:

```
$ lychee --include-fragments hcloud/*.go
  121/121 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                                                                                                                                                                            Issues found in 7 inputs. Find details below.

[hcloud/client.go]:
     [404] https://api.hetzner.cloud/v1 | Rejected status code (this depends on your "accept" configuration): Not Found
     [401] https://api.hetzner.com/v1 | Rejected status code (this depends on your "accept" configuration): Unauthorized

[hcloud/client_test.go]:
   [ERROR] http://api/v1.0///// | Network error: error sending request for url (http://api/v1.0/////) Maybe a certificate error?

[hcloud/schema_test.go]:
   [ERROR] https://osi-preview-openapi-spec-review-storage-bo-w2bwiv.cloud.dev.hc-k8s.de/changelog#2023-07-20-foo-endpoint-is-deprecated | Cannot find fragment

```

(These are all false-positives from examples that need to be ignored)

jooola:
Closes #755

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ err = client.Action.WaitFor(ctx, result.Action)
 
 This release adds support for the new [DNS API](https://docs.hetzner.cloud/reference/cloud#dns).
 
-The DNS API is currently in **beta**, which will likely end on 10 November 2025. After the beta ended, it will no longer be possible to create new zones in the old DNS system. See the [DNS Beta FAQ](https://docs.hetzner.com/networking/dns/faq/beta) for more details.
+The DNS API is currently in **beta**, which will likely end on 10 November 2025. After the beta ended, it will no longer be possible to create new zones in the old DNS system. See the [DNS Beta FAQ](https://docs.hetzner.com/networking/dns/faq/beta/) for more details.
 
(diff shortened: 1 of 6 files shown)


.github/workflows/links.yml
```diff
@@ -0,0 +1,37 @@
+name: Links
+
+on:
+  pull_request:
```
apricote, line 4: is there a really easy way to only look at the changed files for PRs? Don't want to block something just because an unrelated URL broke

apricote: https://lychee.cli.rs/github_action_recipes/pull-requests/ and https://github.com/lycheeverse/lychee-action/issues/238

But that does not at all look like a "really easy way", so lets just merge it as is :)

PR #775, opened 2025-11-25
feat: error helper to remove correlation ID

lukasmetzner:
Helper function to remove any correlation ID from a `hcloud.Error`.

### DIFF
--- a/hcloud/error_test.go
+++ b/hcloud/error_test.go
@@ -154,7 +154,7 @@ func TestStabilizeError(t *testing.T) {
 			name: "error with correlation ID",
 			response: &Response{
 				Response: &http.Response{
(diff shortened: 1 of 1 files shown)


hcloud/error.go
```diff
@@ -157,6 +157,16 @@ func IsError(err error, code ...ErrorCode) bool {
 	return ok && slices.Index(code, apiErr.Code) > -1
 }
 
+// StableError returns an error without any correlation ID.
+func StableError(err error) error {
```
jooola:
```suggestion
func StabilizeError(err error) error {
```

jooola (approved):

hcloud/error.go
lukasmetzner: renamed :+1:

PR #788, opened 2025-12-10
chore: revert Storage Box support no longer experimental

phm07:
This reverts commit daf9cf48e033b234fe09e07fa010f393df6a15bd.
There might need to be breaking changes in the future so it is not safe yet to remove the experimental marking.


jooola (approved):

jooola:
We might need to override the title of the initial PR to remove it from the release notes.

PR #790, opened 2025-12-12
docs: append coding guidelines to maintainer docs

phm07:
This PR adds coding guidelines to our docs.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -20,7 +20,7 @@ Backports are done by [tibdex/backport](https://github.com/tibdex/backport). App
   - Sometimes the API treats an empty value as `nil`, for example an empty string in name updates. There no pointer should be used.
 
 ```go
-type ExampleUpdateOpts struct {
+type ExampleUpdateRequest struct {
     // Name is optional but cannot be empty.
     // It doesn't need to be a pointer, we can represent empty values with "".
     Name string `json:"name,omitempty"`


MAINTAINERS.md
```diff
@@ -9,3 +9,74 @@ All development targets the `main` branch. New releases for the latest major ver
...
+    // It doesn't need to be a pointer, we can represent empty values with "".
+    Name string `json:"name,omitempty"`
+    // Description is optional and can be empty.
+    // We need a pointer to distinguish between an empty value and no value.
+    Description *string `json:"description,omitempty"`
+    // Foo is required, it should not be a pointer and should not be omitted.
+    Foo int `json:"foo"`
+    // Bar is required but nullable. It should be a pointer.
+    Bar *int `json:"bar,omitempty"`
+}
```
jooola:
```suggestion
type ExampleUpdateRequest struct {
    // Name is optional but cannot be empty.
    // It doesn't need to be a pointer, we can represent empty values with "".
    Name string `json:"name,omitempty"`
    // Description is optional and can be empty.
    // We need a pointer to distinguish between an empty value and no value.
    Description *string `json:"description,omitempty"`
    // Foo is required, it should not be a pointer and should not be omitted.
    Foo int `json:"foo"`
    // Bar is required but nullable. It should be a pointer.
    Bar *int `json:"bar,omitempty"`
}
```

`Opts` is for the `hcloud` package, `Request` is for the `schema` package. We must make a difference in this document.

apricote (approved):

PR #791, opened 2025-12-16
feat: list actions for a given resource

jooola:
- Adds `client.<resource_client>.Action.ListFor(ctx, resource, opts)` method to list actions for a resource.
- Adds `client.<resource_client>.Action.AllFor(ctx, resource, opts)` method to list all actions for a resource.

---

### `ResourceActionClient` now has a generic type parameter

The [`ResourceActionClient`](https://pkg.go.dev/github.com/hetznercloud/hcloud-go/v2/hcloud#ResourceActionClient) struct now has a generic type parameter. The type parameter is used to describe the API resource of the Actions client.

The [`IResourceActionClient`](https://pkg.go.dev/github.com/hetznercloud/hcloud-go/v2/hcloud#IResourceActionClient) interface that is generated from the [`ResourceActionClient`](https://pkg.go.dev/github.com/hetznercloud/hcloud-go/v2/hcloud#ResourceActionClient) struct, is also affected by this change.

The API was never intended to be used directly and we did not find any usage of it in public code. So while potentially breaking, the benefits of this change out weighted the risks of doing it.

### DIFF
--- a/hcloud/certificate.go
+++ b/hcloud/certificate.go
@@ -82,11 +82,11 @@ type Certificate struct {
 	UsedBy         []CertificateUsedByRef
 }
 
-func (c *Certificate) pathID() (string, error) {
-	if c.ID == 0 {
-		return "", missingField(c, "ID")
+func (o *Certificate) pathID() (string, error) {
+	if o.ID == 0 {
+		return "", missingField(o, "ID")
 	}
-	return strconv.FormatInt(c.ID, 10), nil
+	return strconv.FormatInt(o.ID, 10), nil
 }
 
 // CertificateCreateResult is the result of creating a certificate.
--- a/hcloud/firewall.go
+++ b/hcloud/firewall.go
@@ -22,11 +22,11 @@ type Firewall struct {
 	AppliedTo []FirewallResource
 }
 
-func (f *Firewall) pathID() (string, error) {
-	if f.ID == 0 {
-		return "", missingField(f, "ID")
+func (o *Firewall) pathID() (string, error) {
+	if o.ID == 0 {
+		return "", missingField(o, "ID")
 	}
-	return strconv.FormatInt(f.ID, 10), nil
+	return strconv.FormatInt(o.ID, 10), nil
 }
 
 // FirewallRule represents a Firewall's rules.
--- a/hcloud/floating_ip.go
+++ b/hcloud/floating_ip.go
@@ -29,11 +29,11 @@ type FloatingIP struct {
 	Name         string
 }
 
-func (f *FloatingIP) pathID() (string, error) {
-	if f.ID == 0 {
-		return "", missingField(f, "ID")
+func (o *FloatingIP) pathID() (string, error) {
+	if o.ID == 0 {
+		return "", missingField(o, "ID")
 	}
-	return strconv.FormatInt(f.ID, 10), nil
+	return strconv.FormatInt(o.ID, 10), nil
 }
 
 // DNSPtrForIP returns the reverse DNS pointer of the IP address.
--- a/hcloud/image.go
+++ b/hcloud/image.go
@@ -35,11 +35,11 @@ type Image struct {
 	Deleted    time.Time
 }
 
-func (image *Image) pathID() (string, error) {
-	if image.ID == 0 {
-		return "", missingField(image, "ID")
+func (o *Image) pathID() (string, error) {
+	if o.ID == 0 {
+		return "", missingField(o, "ID")
 	}
-	return strconv.FormatInt(image.ID, 10), nil
+	return strconv.FormatInt(o.ID, 10), nil
 }
 
 // IsDeprecated returns whether the image is deprecated.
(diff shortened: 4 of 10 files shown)


hcloud/action_test.go
```diff
@@ -330,3 +332,130 @@ func TestResourceActionClientAll(t *testing.T) {
 		t.Errorf("unexpected actions")
 	}
 }
+
+func TestResourceActionClientListFor(t *testing.T) {
```
lukasmetzner, line 337: Is it worth adding a small test example for zones?

jooola: I added one.

hcloud/action.go
```diff
@@ -194,3 +194,29 @@ func (c *ResourceActionClient) All(ctx context.Context, opts ActionListOpts) ([]
...
 	})
 }
+
+// ListFor returns a paginated list of actions for the given Resource.
+//
+// Please note that filters specified in opts are not taken into account
+// when their value corresponds to their zero value or when they are empty.
+func (c *ResourceActionClient) ListFor(ctx context.Context, resourceID any, opts ActionListOpts) ([]*Action, *Response, error) {
+	opPath := c.getBaseURL() + "/%v/actions?%s"
+	ctx = ctxutil.SetOpPath(ctx, opPath)
```
jooola: I wish we could use a generic here, but that would be a breaking change.

Or having a resource path id function (using an interface) could also be an option, but would require all resource to implement one.

jooola: I implement partially using an interface, let me know what you think.

lukasmetzner: Looks good to me.

Should we maybe put `ResourceWithPathID` in a different file? Maybe we want to extend this in the future with other interfaces? `ResourceWithID`, `ResourceWithIDOrName`?

phm07:
Is there maybe a way to ensure that resources implement the `ResourceWithPathID` interface? I could see this being accidentally missed for future resources.

jooola:
> Is there maybe a way to ensure that resources implement the ResourceWithPathID interface? I could see this being accidentally missed for future resources.

We can have a check that the resource implements the interface. Tests would be the best way to check this.

hcloud/resource.go
```diff
@@ -5,3 +5,19 @@ type Resource struct {
...
+var _ SupportsActions = &Firewall{}
+var _ SupportsActions = &FloatingIP{}
+var _ SupportsActions = &Image{}
+var _ SupportsActions = &LoadBalancer{}
+var _ SupportsActions = &Network{}
+var _ SupportsActions = &PrimaryIP{}
+var _ SupportsActions = &Server{}
+var _ SupportsActions = &Volume{}
+var _ SupportsActions = &Zone{}
+var _ SupportsActions = &StorageBox{}
```
apricote: Could you move this close to where we add the `ActionResourceClient` to each resource?

I feel having this in a central place risks a mismatch between `$Resource` structs with a `Action ActionResourceClient` property and them implementing this interface.

hcloud/action.go
apricote: @jooola could you explain why you reverted the interface usage in https://github.com/hetznercloud/hcloud-go/pull/791/changes/c29ea99dee9b47156c92ad40af3f4a3715484d12?

apricote:
I think the benefits of adding a generic type parameter to `ResourceActionClient` outweigh the actual impact of it. I did not find any project outside of the CLI that would need to be updated in my search on [Sourcegraph](https://sourcegraph.com/search?q=lang:go+-file:vendor/+-repo:github.com/hetznercloud/hcloud-go+-repo:github.com/kubernetes/autoscaler+content:%22hetznercloud/hcloud-go%22+-repo:github.com/hetznercloud/cli+-repo:github.com/woodpecker-ci/autoscaler+-repo:github.com/quizzicalinst/autoscaler+count:all+%28ResourceActionClient+OR+IResourceActionClient%29&patternType=keyword&sm=0) and [GitHub](https://github.com/search?q=-org%3Ahetznercloud+language%3Ago+%28ResourceActionClient+OR+IResourceActionClient%29++NOT+is%3Aarchived+NOT+is%3Afork+NOT+path%3A%22cluster-autoscaler%2Fcloudprovider%2Fhetzner%2Fhcloud-go%22+%22hcloud%22&type=code).

We should still highlight this in the release notes. The usage of the private interface as a constraint for the type parameter should be evaluated in the CLI.

hcloud/action.go
```diff
@@ -201,3 +207,35 @@ func (c *ResourceActionClient) All(ctx context.Context, opts ActionListOpts) ([]
 		return c.List(ctx, opts)
 	})
 }
+
+type actionSupporter interface {
+	pathID() string
+}
```
jooola, line 207: I first implemented this without error handling, but this might shoot back at us.

Should we maybe add an error ? 

```suggestion
type actionSupporter interface {
	pathID() (string, error)
}
```

jooola: I'll implement this.

jooola:
Just tried this branch in the CLI project, and no change are required to make this work.

PR #793, opened 2025-12-16
feat: deprecate datacenter in `primary ips` and `servers`

jooola:
See https://docs.hetzner.cloud/changelog#2025-12-16-phasing-out-datacenters

### DIFF
--- a/hcloud/server_test.go
+++ b/hcloud/server_test.go
@@ -2,7 +2,6 @@ package hcloud
 
 import (
 	"context"
-	_ "embed"
 	"encoding/json"
 	"fmt"
 	"net"


hcloud/schema/server.go
```diff
@@ -18,14 +18,19 @@ type Server struct {
...
 	Labels          map[string]string  `json:"labels"`
 	Volumes         []int64            `json:"volumes"`
 	PrimaryDiskSize int                `json:"primary_disk_size"`
 	PlacementGroup  *PlacementGroup    `json:"placement_group"`
 	LoadBalancers   []int64            `json:"load_balancers"`
+
+	// Deprecated: [Server.Datacenter] is deprecated and will be removed after 1 July 2026.
+	// Use [Server.Location] instead.
+	// See https://docs.hetzner.cloud/changelog#2025-12-16-phasing-out-datacenters
+	Datacenter Datacenter `json:"datacenter"`
```
phm07: Maybe we should make it a pointer and add an `omitempty` tag so that the `datacenter` property isn't included in the schema anymore when it's removed from the API.

jooola: Good point, fixed.

apricote (approved):

PR #794, opened 2025-12-18
feat: default to max per page value in `AllWithOpts`

### LINKED ISSUE #792: Set maxium per page in AllWithOpts(...) instead of All(...)
### TL;DR

The `All(...)` calls sets the per page value to the maximum of the endpoint. This is not the case for `AllWithOpts(...)` which uses the default per page value.

### Expected behavior

Set the per page value to the maximum in each `AllWithOpts` function, if the user did not provide a value (0).

### Observed behavior

`AllWithOpts` does not use the maximum per page value that is used in `All`.

### Minimal working example

_No response_

### Log output

```plain text

```

### Additional information

_No response_

jooola:
Fixes #792

### DIFF
--- a/hcloud/action.go
+++ b/hcloud/action.go
@@ -128,13 +128,17 @@ func (c *ActionClient) List(ctx context.Context, opts ActionListOpts) ([]*Action
 //
 // Deprecated: It is required to pass in a list of IDs since 30 January 2025. Please use [ActionClient.AllWithOpts] instead.
 func (c *ActionClient) All(ctx context.Context) ([]*Action, error) {
-	return c.action.All(ctx, ActionListOpts{ListOpts: ListOpts{PerPage: 50}})
+	return c.action.All(ctx, ActionListOpts{})
 }
 
 // AllWithOpts returns all actions for the given options.
 //
 // It is required to set [ActionListOpts.ID]. Any other fields set in the opts are ignored.
 func (c *ActionClient) AllWithOpts(ctx context.Context, opts ActionListOpts) ([]*Action, error) {
+	if opts.ListOpts.PerPage == 0 {
+		// Do not send unused per page param
+		opts.ListOpts.PerPage = -1
+	}
 	return c.action.All(ctx, opts)
 }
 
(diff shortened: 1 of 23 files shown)


phm07 (approved):
Maybe something like `const MaxItemsPerPage = 50` and then using that constant instead would be nice to eliminate the magic number. But works without that as well

hcloud/action.go
```diff
@@ -128,13 +128,17 @@ func (c *ActionClient) List(ctx context.Context, opts ActionListOpts) ([]*Action
...
+	return c.action.All(ctx, ActionListOpts{})
 }
 
 // AllWithOpts returns all actions for the given options.
 //
 // It is required to set [ActionListOpts.ID]. Any other fields set in the opts are ignored.
 func (c *ActionClient) AllWithOpts(ctx context.Context, opts ActionListOpts) ([]*Action, error) {
+	if opts.ListOpts.PerPage == 0 {
+		// Do not send unused per page param
+		opts.ListOpts.PerPage = -1
```
apricote, line 140: Would this not send `?per_page=-1` in the request?

jooola: No we have:
```
	if l.Page > 0 {
		vals.Add("page", strconv.Itoa(l.Page))
	}
```

apricote: Why is it necessary to change `0` to `-1` then? Both should not be added to the query string

https://github.com/hetznercloud/hcloud-go/blob/6afa559db0976c24aaf05951de41231f0aba629f/hcloud/client.go#L449-L451

jooola: It goes like this:

https://github.com/hetznercloud/hcloud-go/blob/2f15ac50e7ee9e3d5e0f31eafd5705ffdaad4351/hcloud/action.go#L137-L141

https://github.com/hetznercloud/hcloud-go/blob/2f15ac50e7ee9e3d5e0f31eafd5705ffdaad4351/hcloud/action.go#L195-L198

https://github.com/hetznercloud/hcloud-go/blob/2f15ac50e7ee9e3d5e0f31eafd5705ffdaad4351/hcloud/client.go#L449-L451

Note that we use the resource action client for the "main" action client.

apricote: Ah yes, as `ActionClient.AllWithOpts` only supports the ID query now. Resolving..

PR #838, opened 2026-04-23
Add proxy configuration

1fexd:
Allow configuration of a proxy for the http client


jooola:
Hey @1fexd,
You could provide your own http client directly with: https://pkg.go.dev/github.com/hetznercloud/hcloud-go/v2@v2.38.0/hcloud#WithHTTPClient

With that in mind, do you really think a specialized option for a custom transport proxy is really needed?

PR #843, opened 2026-04-29
refactor: generate top level schema conversion functions using gowrap

phm07:
Before #343, we used to manually write schema conversion functions in the top level `hcloud` package. Since switching to goverter, these functions now wrap the `converter` interface, since goverter can only generate a struct that implements the  interface. This means we have to manually add top level functions, which is tedious and error prone.
This PR utilizes https://github.com/hexdigest/gowrap to run the `converter` interface through a template that generates `schema.go` automatically.


hcloud/schema.go
```diff
@@ -1,572 +1,741 @@
...
 // This file provides converter functions to convert models in the
 // schema package to models in the hcloud package and vice versa.
 
 var c converter
 
-// ActionFromSchema converts a schema.Action to an Action.
-func ActionFromSchema(s schema.Action) *Action {
-	return c.ActionFromSchema(s)
+// ActionFromSchema converts schema.Action to *Action.
+func ActionFromSchema(a1 schema.Action) (ap1 *Action) {
```
jooola: Could we use something a stable argument name here? e.g. `s` for any schema? And maybe no return value variable if not needed?

hcloud/schema.go
```diff
@@ -1,572 +1,741 @@
...
+// template: schema.tmpl
+// gowrap: http://github.com/hexdigest/gowrap
+
 package hcloud
 
 import (
 	"github.com/hetznercloud/hcloud-go/v2/hcloud/schema"
 )
 
+//go:generate go run github.com/hexdigest/gowrap/cmd/gowrap gen -g -p . -i converter -t schema.tmpl -o schema.go
```
jooola: Hmm, so this file is now generated, so should we prefix its filename with `zz_`? This would might need moving the generate function and the existing code somewhere else. Since we delete all `zz_*` files in the CI.

phm07: The conversion functions are referenced in other source files, so compilation would fail. This is a problem because some tools (like goverter) need a compiling codebase to generate code. I agree that it is maybe not that pretty but I don't think there is a way around it.

Edit: Nevermind, `//go:generate` seems to respect order. So it would be possible like it is done in 1f3173410dd0cf94c7fa39e20218f3c429186d38

hcloud/schema.go
phm07: Yes, we could build `{{$method.Declaration}}` by hand. I can write something up.

phm07: See 3a26c5df21d0bdaf229f696d80c87128882df7a9

jooola (approved):

PR #849, opened 2026-05-08
docs: add hints for `go fix` tool

phm07:
This PR adds `//go:fix inline` hints to compatible deprecated functions. This makes it possible for the `go fix` tool to automatically replace deprecated members with their recommended alternative.

### Example
Before
```go
func foo() {
	client := hcloud.NewClient(hcloud.WithBackoffFunc(hcloud.ConstantBackoff(5 * time.Second)))
	client = hcloud.NewClient(hcloud.WithPollBackoffFunc(hcloud.ExponentialBackoff(2.0, 2)))
	client = hcloud.NewClient(hcloud.WithPollInterval(5 * time.Second))

	_ = hcloud.String("abc")
	_ = hcloud.Duration(time.Second)
	_ = hcloud.Int(123)
	_ = hcloud.Bool(true)

	var err hcloud.Error
	if err.Code == hcloud.ErrorCodeLimitReached {
		/* ... */
	}

	var (
		_ hcloud.PrimaryIPAssignResult
		_ hcloud.PrimaryIPChangeDNSPtrResult
		_ hcloud.PrimaryIPChangeProtectionResult
	)
}
```

After `go fix`
```go
func foo() {
	client := hcloud.NewClient(hcloud.WithRetryOpts(hcloud.RetryOpts{BackoffFunc: hcloud.ConstantBackoff(5 * time.Second)}))
	client = hcloud.NewClient(hcloud.WithPollOpts(hcloud.PollOpts{BackoffFunc: hcloud.ExponentialBackoff(2.0, 2)}))
	client = hcloud.NewClient(hcloud.WithPollOpts(hcloud.PollOpts{BackoffFunc: hcloud.ConstantBackoff(5 * time.Second)}))

	_ = hcloud.Ptr("abc")
	_ = hcloud.Ptr(time.Second)
	_ = hcloud.Ptr(123)
	_ = hcloud.Ptr(true)

	var err hcloud.Error
	if err.Code == hcloud.ErrorCodeRateLimitExceeded {
		/* ... */
	}

	var (
		_ schema.PrimaryIPActionAssignResponse
		_ schema.PrimaryIPActionChangeDNSPtrResponse
		_ schema.PrimaryIPActionChangeProtectionResponse
	)
}
```

### DIFF
--- a/hcloud/action.go
+++ b/hcloud/action.go
@@ -134,10 +134,8 @@ func (c *ActionClient) List(ctx context.Context, opts ActionListOpts) ([]*Action
 // All returns all actions.
 //
 // Deprecated: It is required to pass in a list of IDs since 30 January 2025. Please use [ActionClient.AllWithOpts] instead.
-//
-//go:fix inline
 func (c *ActionClient) All(ctx context.Context) ([]*Action, error) {
-	return c.AllWithOpts(ctx, ActionListOpts{ID: []int64{}})
+	return c.action.All(ctx, ActionListOpts{})
 }
 
 // AllWithOpts returns all actions for the given options.
--- a/hcloud/action.go
+++ b/hcloud/action.go
@@ -134,8 +134,10 @@ func (c *ActionClient) List(ctx context.Context, opts ActionListOpts) ([]*Action
 // All returns all actions.
 //
 // Deprecated: It is required to pass in a list of IDs since 30 January 2025. Please use [ActionClient.AllWithOpts] instead.
+//
+//go:fix inline
 func (c *ActionClient) All(ctx context.Context) ([]*Action, error) {
-	return c.action.All(ctx, ActionListOpts{})
+	return c.AllWithOpts(ctx, ActionListOpts{ID: []int64{}})
 }
 
 // AllWithOpts returns all actions for the given options.
--- a/hcloud/helper.go
+++ b/hcloud/helper.go
@@ -34,7 +34,3 @@ func Bool(b bool) *bool { return Ptr(b) }
 //
 //go:fix inline
 func Duration(d time.Duration) *time.Duration { return Ptr(d) }
-
-func IgnoreError[T any](t T, _ error) T {
-	return t
-}
--- a/hcloud/helper.go
+++ b/hcloud/helper.go
@@ -10,19 +10,31 @@ func Ptr[T any](p T) *T {
 // String returns a pointer to the passed string s.
 //
 // Deprecated: Use [Ptr] instead.
+//
+//go:fix inline
 func String(s string) *string { return Ptr(s) }
 
 // Int returns a pointer to the passed integer i.
 //
 // Deprecated: Use [Ptr] instead.
+//
+//go:fix inline
 func Int(i int) *int { return Ptr(i) }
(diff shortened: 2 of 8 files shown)


hcloud/action.go
```diff
@@ -134,8 +134,10 @@ func (c *ActionClient) List(ctx context.Context, opts ActionListOpts) ([]*Action
 // All returns all actions.
 //
 // Deprecated: It is required to pass in a list of IDs since 30 January 2025. Please use [ActionClient.AllWithOpts] instead.
+//
+//go:fix inline
 func (c *ActionClient) All(ctx context.Context) ([]*Action, error) {
-	return c.action.All(ctx, ActionListOpts{})
+	return c.AllWithOpts(ctx, ActionListOpts{ID: []int64{}})
```
phm07: I added `ID: []int64{}` explicitly here to make clear that a list of IDs is expected. This function should have been broken for over a year anyway.

hcloud/helper.go
```diff
@@ -10,19 +10,31 @@ func Ptr[T any](p T) *T {
...
 func Bool(b bool) *bool { return Ptr(b) }
 
 // Duration returns a pointer to the passed time.Duration d.
 //
 // Deprecated: Use [Ptr] instead.
+//
+//go:fix inline
 func Duration(d time.Duration) *time.Duration { return Ptr(d) }
+
+func IgnoreError[T any](t T, _ error) T {
```
phm07: This helper is necessary to make it possible to inline (*FloatingIP).DNSPtrForIP. If we don't want to export it we could consider moving it to exp/ (though there would be an import cycle if we moved it to `errutil`) or removing it entirely.

jooola: Hmm, I am not sure this is a good idea. We should probably not officially suggest ignoring errors. I'd rather leave this function out and require manual intervention from the users.

hcloud/action.go
jooola: Hmm, there is no real replacement for this function, the feature has been removed, so I think we should only have it deprecated. Fixing this requires manual intervention.

Otherwise we replace the deprecated function with a function that is not deprecated but also not working, and we remove the information from the users that the feature is removed.

jooola (commented):
Great idea, just a few nits.

hcloud/helper.go
```diff
@@ -10,19 +10,31 @@ func Ptr[T any](p T) *T {
 // String returns a pointer to the passed string s.
 //
 // Deprecated: Use [Ptr] instead.
+//
+//go:fix inline
 func String(s string) *string { return Ptr(s) }
```
jooola, line 15: Hmm, we will soon be able to suggest new(d), is this really worth changing now?

apricote: `go fix` is smart enough to detect that the function `Ptr()` is equivalent to `new(d)` and replaces that correctly.

@phm07 mentioned to me that `go fix` also works recursively, so if the user is on Go 1.26, and they still use `hcloud.String()` it will be replaced with `new()`.

jooola: Nice :)

jooola (approved):

PR #852, opened 2026-05-19
feat(metadata): add context-aware Client methods

### LINKED ISSUE #851: metadata: add context.Context support to Client methods
### TL;DR

The metadata `Client` methods (`Hostname`, `InstanceID`, `Region`, `AvailabilityZone`, etc.) do not accept a `context.Context`. Internally, `Client.get` creates its own `context.Background()`, making it impossible for callers to propagate deadlines or cancellations. We use this package in [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) and [opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib), where the `resource.Detector` interface receives a `context.Context` that cannot be forwarded.

### Expected behavior

Each public `Client` method (or at minimum the underlying get method) should accept and forward a `context.Context` so that callers can control the request lifecycle. For example: `func (c *Client) Hostname(ctx context.Context) (string, error)`.

### Observed behavior

`Client.get` hardcodes `context.Background()` on every HTTP request. Any context passed by the caller is silently ignored, preventing cancellation, deadline enforcement, or trace propagation.

### Minimal working example

_No response_

### Log output

```plain text

```

### Additional information

_No response_

c-tonneslan:
The metadata `Client` methods build their HTTP requests with `context.Background()` internally, so callers can't propagate deadlines or cancellations. This came up while wiring the Hetzner detector in opentelemetry-go-contrib: `resource.Detector` receives a context and has no way to forward it, so the workaround in [open-telemetry/opentelemetry-go-contrib#8999](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8999) is a goroutine + `select` on `ctx.Done()`. The contrib maintainers said they'd rather wait for the fix to live here.

This adds `XxxWithContext` variants for each public method (`HostnameWithContext`, `InstanceIDWithContext`, `PublicIPv4WithContext`, `RegionWithContext`, `AvailabilityZoneWithContext`, `PrivateNetworksWithContext`, and `IsHcloudServerWithContext`) and threads a `ctx` parameter through the internal `get`. The original no-arg methods keep working, they just wrap the new ones with `context.Background()`. I marked them `Deprecated:` to nudge new code over, matching the style used in `action.go` and elsewhere, but happy to drop those if you'd rather keep the existing API undeprecated and revisit in a major version.

Tests cover the new happy path, a cancelled-context case on `get`, and a deadline-exceeded case on `HostnameWithContext`.

Closes #851.

### DIFF
--- a/hcloud/metadata/client.go
+++ b/hcloud/metadata/client.go
@@ -81,8 +81,8 @@ func NewClient(options ...ClientOption) *Client {
 }
 
 // get executes an HTTP request against the API.
-func (c *Client) get(path string) (string, error) {
-	ctx := ctxutil.SetOpPath(context.Background(), path)
+func (c *Client) get(ctx context.Context, path string) (string, error) {
+	ctx = ctxutil.SetOpPath(ctx, path)
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.endpoint+path, http.NoBody)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


jooola (approved):
Thanks for the contribution 🚀

apricote:
We could use the new `//go:fix inline` hints (cf. #849) on the old methods and use `context.TODO()` instead of background. Then any consumers will see the TODO context possibly hook it up properly.

```diff
+//go:fix inline
 func (c *Client) IsHcloudServer() bool {
- 	return c.IsHcloudServerWithContext(context.Background())
+ 	return c.IsHcloudServerWithContext(context.TODO())
 }

PR #861, opened 2026-05-29
refactor: replace interface{} with any (Go 1.18+)

MD-Mushfiqur123:
Replace `interface{}` with `any` (the Go 1.18+ type alias) in comments for consistency with modern Go idioms.


jooola:
Hey,
The concept of "interfaces" being mentioned in those comment is still accurate. Replacing the word "interface" with "any" here does not seem to help to better understand the comment.

Thanks for the effort though.

PR #866, opened 2026-06-03
feat(metadata): add user agent header

jooola:
- Move version to dedicated internal package, to break a cyclic import.
- Add a user agent header to the metadata client.
- Extract user agent utility to an internal package.


lukasmetzner (approved):
Looking good. Nice feature 🚀

PR #867, opened 2026-06-04
feat: deprecate datacenters resource

jooola:
The API endpoints `GET /v1/datacenters` and `GET /v1/datacenters/{id}` are now deprecated and will be removed after 1 Oct. 2026. After this date, requests to these endpoints will return `HTTP 410 Gone`.

See https://docs.hetzner.cloud/changelog#2026-06-02-datacenters-deprecated.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -69,6 +69,9 @@ linters:
         text: "unused-parameter: parameter '(w|r)' seems to be unused, consider removing or renaming it as _"
       - linters:
           - staticcheck
+        text: "\\[Datacenter\\] is deprecated"
+      - linters:
+          - staticcheck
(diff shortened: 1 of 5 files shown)


apricote (approved):
Do we also want to deprecate the [`Datacenter` field in the `hcloud.Client` struct](https://github.com/hetznercloud/hcloud-go/blob/0bbc6711d1f2a88634ef397415acc89c7d7a5f88/hcloud/client.go#L106)?

jooola:
> Do we also want to deprecate the [Datacenter field in the hcloud.Client struct](https://github.com/hetznercloud/hcloud-go/blob/0bbc6711d1f2a88634ef397415acc89c7d7a5f88/hcloud/client.go#L106)?

Yes indeed.

PR #890, opened 2026-07-06
feat: remove datacenter property from server and primary_ip

lukasmetzner:
Remove the `datacenter` property from Server and Primary IP resource where possible. The property was deprecated ~6 months ago, and because the property is removed from the API, removing it from our code is not considered a breaking change.

Changelog entry: https://docs.hetzner.cloud/changelog#2026-07-01-removing-datacenters

Assisted-by: Claude Opus 4.8


jooola:
Could you please add a changelog link in the pr description?

jooola (approved):

PR #921, opened 2026-09-16
feat: propose block volume snapshot API

eugene-harold-krabs:
## Proposal

Add the client contract needed for native Hetzner Cloud block-volume snapshots.
This is intentionally a draft because the public Cloud API endpoints do not
exist yet; it gives the API and CSI teams concrete request/response types to
review before server-side shipment.

### Proposed Cloud API

- `POST /volumes/{id}/actions/create_snapshot` accepts `name`, optional
  `description`, and optional `labels`; it returns `snapshot` plus an async
  `action`.
- `GET /volume_snapshots/{id}` returns one snapshot.
- `GET /volume_snapshots` uses normal pagination and supports `name`, `volume`,
  `status`, `label_selector`, and `sort` filters.
- `DELETE /volume_snapshots/{id}` returns an async deletion `action`.
- `POST /volumes` accepts an optional `snapshot` ID. `size` must be at least the
  source snapshot size and the restored volume stays in the snapshot location.

The proposed snapshot resource exposes `id`, `name`, `description`, `status`,
source `volume`, `location`, `size`, `labels`, and `created`. Creation is
durable when its action succeeds and the resource reaches `available`.

### Client surface

- Adds typed create/get/list/delete methods to `VolumeClient`.
- Adds `VolumeCreateOpts.Snapshot` for restore.
- Regenerates `IVolumeClient`.
- Documents the behavior and tests every proposed request path and restore
  payload against the mock API.

## Validation

- `go generate ./hcloud`
- `go test ./...`

## Tracking

- hetznercloud/csi-driver#849
- https://gitlab.com/xrow-public/infrastructure/-/work_items/46

The companion CSI proposal will compile against this branch and implement the
CSI `CREATE_DELETE_SNAPSHOT` contract end to end.


jooola:
Closing, see https://github.com/hetznercloud/csi-driver/issues/849#issuecomment-5694717485.

